PR #17, opened 2023-10-04
Feature/pdct 451 add summary endpoint

katybaulch:
# Description

- Replaced dummy response with analytics summary data using count queries on the document, family, and collection services.
- "n_events" is currently set to 0 as we have not implemented events in the admin backend yet (see [here](https://github.com/climatepolicyradar/navigator-backend/blob/ddebbd17f6b62cf7909e6e4c575285b8b00a41b2/app/db/models/law_policy/family.py#L218). When we implement events in the admin backend, we should update how "n_events" is calculated.

[Link to Linear ticket PDCT-451](https://linear.app/climate-policy-radar/issue/PDCT-451/add-summaries-endpoint)

## Type of change

- [ ] Bug fix
- [x ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Tested locally. Actual tests to come.

## Reviewer Checklist

- [ ] The PR represents a single feature (small driveby fixes are also ok)
- [ ] No FIXMEs remain

### DIFF
--- a/app/service/collection.py
+++ b/app/service/collection.py
@@ -135,3 +135,18 @@ def delete(import_id: str, db: Session = db_session.get_db()) -> bool:
     """
     id.validate(import_id)
     return collection_repo.delete(db, import_id)
+
+
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
+def count() -> Optional[int]:
+    """
+    Gets a count of collections from the repository.
+
+    :return Optional[int]: The number of collections in the repository or none.
+    """
+    try:
+        with db_session.get_db() as db:
+            return collection_repo.count(db)
+    except exc.SQLAlchemyError as e:
+        _LOGGER.error(e)
+        raise RepositoryError(str(e))
--- a/app/service/document.py
+++ b/app/service/document.py
@@ -3,8 +3,8 @@ from typing import Optional, Tuple
 
 from pydantic import ConfigDict, validate_call
 from app.clients.aws.client import get_s3_client
-from app.errors import RepositoryError
-from app.model.document import DocumentReadDTO, DocumentWriteDTO
+from app.errors import RepositoryError, ValidationError
+from app.model.document import DocumentCreateDTO, DocumentReadDTO, DocumentWriteDTO
 import app.repository.document as document_repo
 import app.repository.document_file as file_repo
 import app.clients.db.session as db_session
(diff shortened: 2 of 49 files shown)


diversemix:
Amazing! This is exactly how it should be done :+1: :smiley_cat:

app/service/collection.py
```diff
@@ -135,3 +135,18 @@ def delete(import_id: str, db: Session = db_session.get_db()) -> bool:
     """
     id.validate(import_id)
     return collection_repo.delete(db, import_id)
+
+
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
```
diversemix, line 140: Looks like the analytics service uses the repo directly, so is there a need/use of this function?

app/service/document.py
```diff
@@ -136,3 +136,18 @@ def delete(import_id: str, db: Session = db_session.get_db()) -> bool:
     """
     id.validate(import_id)
     return document_repo.delete(db, import_id)
+
+
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
```
diversemix, line 141: Looks like the analytics service uses the repo directly, so is there a need/use of this function?

app/service/family.py
```diff
@@ -145,3 +145,19 @@ def delete(import_id: str, db: Session = db_session.get_db()) -> bool:
         msg = f"Unable to delete family {import_id}"
         _LOGGER.exception(msg)
         raise RepositoryError(msg)
+    return family_repo.delete(db, import_id)
+
+
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
```
diversemix, line 151: Looks like the analytics service uses the repo directly, so is there a need/use of this function?

app/service/collection.py
katybaulch: Updated analytics service to use the services instead of repositories.

app/service/family.py
katybaulch: Updated analytics service to use the services instead of repositories.

app/service/document.py
katybaulch: Updated analytics service to use the services instead of repositories.

unit_tests/helpers/analytics.py
```diff
@@ -0,0 +1,18 @@
+from app.model.analytics import SummaryDTO
+
+# ANALYTICS SUMMARY
+EXPECTED_NUM_DOCUMENTS = 33
+EXPECTED_NUM_FAMILIES = 22
+EXPECTED_NUM_COLLECTIONS = 11
+EXPECTED_NUM_EVENTS = 0
+
+
+def create_summary_dto(
```
diversemix, line 10: Probably don't need this - but its nice for symmetry :+1:

katybaulch: I thought the same! Depends how we extend the analytics service going forwards I suppose.

PR #39, opened 2023-10-25
Added integration tests for login edge cases.

katybaulch:
Added integration tests for user login.

### DIFF
--- a/integration_tests/setup_db.py
+++ b/integration_tests/setup_db.py
@@ -210,11 +210,12 @@ def _add_app_user(
     email: str,
     name: str,
     org_id,
+    password: str = "",
     is_active: bool = True,
     is_super: bool = False,
 ):
     test_db.add(
-        AppUser(email=email, name=name, hashed_password="", is_superuser=is_super)
+        AppUser(email=email, name=name, hashed_password=password, is_superuser=is_super)
     )
     test_db.flush()
     test_db.add(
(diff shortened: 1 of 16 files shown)


integration_tests/setup_db.py
```diff
@@ -253,10 +255,14 @@ def _setup_organisation(test_db: Session) -> int:
         "Test",
         org.id,
         "$2b$12$XXMr7xoEY2fzNiMR3hq.PeJBUUchJyiTfJP.Rt2eq9hsPzt9SXzFC",
-        is_active=True,
     )
+    _add_app_user(test_db, "test1@cpr.org", "TestInactive", org.id, is_active=False)
```
diversemix, line 259: Doesn't this user also have an empty hashed password?

integration_tests/setup_db.py
```diff
@@ -253,10 +255,14 @@ def _setup_organisation(test_db: Session) -> int:
...
+    _add_app_user(
+        test_db, "test2@cpr.org", "TestHashedPassEmpty", org.id, hashed_pass=""
+    )
+    _add_app_user(test_db, "test3@cpr.org", "TestPassMismatch", org.id, hashed_pass="")
```
diversemix, line 263: And this one also has an empty hash?

integration_tests/setup_db.py
katybaulch: Technically, we test whether the user is active or not before we check the validity of their password against the hash.

diversemix (approved):
Sweet :candy:

PR #46, opened 2023-11-01
Validate variant

katybaulch:
# Description

Added validation of document variant name in document service and tests.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] The PR represents a single feature (small driveby fixes are also ok)
- [ ] No FIXMEs remain

### DIFF
--- a/app/service/document.py
+++ b/app/service/document.py
@@ -92,6 +92,9 @@ def update(
     """
     validate_import_id(import_id)
 
+    if document.variant_name == "":
+        raise ValidationError("Variant name is empty")
+
     try:
         if document_repo.update(db, import_id, document):
             db.commit()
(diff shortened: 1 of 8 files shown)


app/service/document.py
```diff
@@ -92,6 +92,9 @@ def update(
     """
     validate_import_id(import_id)
 
+    if document.variant_name == "":
```
diversemix, line 95: Sorry - this is my bad, but we also need to check that `variant_name` appears in the db (see https://github.com/climatepolicyradar/navigator-admin-backend/blob/355a3dacba85b2d11c0f8571816ff803e3c98a2b/app/clients/db/models/law_policy/family.py#L23 ) ... happy this is done as a follow-up

diversemix (approved):
lgtm - note my comment - we will also need to check the value for `variant_name` too

PR #52, opened 2023-12-05
Feature/pdct 633 add timeout exception to search api

katybaulch:
# Description

Please include:

- Added timeout error handling
- Added max_results query param and set default to 500
- Updated family search to return 200 with empty response instead of 404 when no matching families found
- Linear ticket [PDCT-633](https://linear.app/climate-policy-radar/issue/PDCT-633/add-timeout-exception-to-search-api)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

Unit & integration tests added.

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/family.py
+++ b/app/api/api_v1/routers/family.py
@@ -87,8 +87,13 @@ async def search_family(request: Request) -> list[FamilyReadDTO]:
     try:
         families = family_service.search(query_params)
     except ValidationError as e:
-        _LOGGER.error(e.message)
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=e.message)
+    except TimeoutError:
+        msg = "Request timed out fetching matching families. Try adjusting your query."
+        raise HTTPException(
+            status_code=status.HTTP_408_REQUEST_TIMEOUT,
+            detail=msg,
+        )
 
     if len(families) == 0:
         _LOGGER.info(f"Families not found for terms: {query_params}")
--- a/app/api/api_v1/routers/family.py
+++ b/app/api/api_v1/routers/family.py
@@ -7,11 +7,13 @@ layer would just pass through directly to the repo. So the approach
 implemented directly accesses the "repository" layer.
 """
 import logging
-from fastapi import APIRouter, HTTPException, Request, status
-from app.errors import RepositoryError, ValidationError
+from typing import Union
 
-from app.model.family import FamilyCreateDTO, FamilyReadDTO, FamilyWriteDTO
+from fastapi import APIRouter, HTTPException, Request, status
+
 import app.service.family as family_service
+from app.errors import RepositoryError, ValidationError
+from app.model.family import FamilyCreateDTO, FamilyReadDTO, FamilyWriteDTO
 
 families_router = r = APIRouter()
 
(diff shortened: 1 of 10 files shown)


app/api/api_v1/routers/family.py
```diff
@@ -78,19 +80,24 @@ async def search_family(request: Request) -> list[FamilyReadDTO]:
...
+    }
 
     try:
         families = family_service.search(query_params)
     except ValidationError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=e.message)
-
-    if len(families) == 0:
+    except TimeoutError:
+        msg = "Request timed out fetching matching families. Try adjusting your query."
```
diversemix, line 92: Might be a good idea to log something here too, just so we know in the logs the timeout has occured

app/repository/family.py
```diff
@@ -179,19 +181,19 @@ def search(db: Session, query_params: dict[str, str]) -> list[FamilyReadDTO]:
         search.append(Family.description.ilike(term))
 
     if "geography" in query_params.keys():
-        term = query_params["geography"]
+        term = cast(str, query_params["geography"])
```
diversemix, line 184: how did we get away without these casts before :man_shrugging: ... nice tidy :smile:

app/service/family.py
```diff
@@ -58,7 +58,7 @@ def all() -> list[FamilyReadDTO]:
 
 
 @validate_call(config=ConfigDict(arbitrary_types_allowed=True))
-def search(query_params: dict[str, str]) -> list[FamilyReadDTO]:
+def search(query_params: dict[str, Union[str, int]]) -> list[FamilyReadDTO]:
```
diversemix, line 61: So my preference is that the responsibility for processing/parsing the query params lies within the router. This then allows the service and the repo to be more rigorously typed, and removes the need for them to understand what and how params are sent over HTTP ... what do you think?

unit_tests/mocks/services/family_service.py
```diff
@@ -23,10 +29,23 @@ def mock_get_family(import_id: str) -> Optional[FamilyReadDTO]:
             return create_family_dto(import_id)
 
     def mock_search_families(q_params: dict) -> list[FamilyReadDTO]:
+        VALID_PARAMS = [
```
diversemix, line 32: Hmmm, would it be better to import `VALID_PARAMS` (if we make them accessible? Otherwise the test for valid params doesn't feel like its testing the actual code :thinking:

diversemix (commented):
Would be good to chat, as it may also be an ideal to set the `statement_timeout` when we call `create_engine` [here](https://github.com/climatepolicyradar/navigator-admin-backend/blob/2913f2e23b17a13ff84a24b99fb65718f695a0ce/app/clients/db/session.py#L8) to something like 10s to improve the user experience.  See https://stackoverflow.com/a/44936982

app/api/api_v1/routers/family.py
katybaulch: Great shout!

katybaulch: Addressed. Link to docs describing how we've configured this here https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CONNECT-TIMEOUT

katybaulch:
> Would be good to chat, as it may also be an ideal to set the `statement_timeout` when we call `create_engine` [here](https://github.com/climatepolicyradar/navigator-admin-backend/blob/2913f2e23b17a13ff84a24b99fb65718f695a0ce/app/clients/db/session.py#L8) to something like 10s to improve the user experience. See https://stackoverflow.com/a/44936982

Addressed. Link to docs describing how we've configured this here https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CONNECT-TIMEOUT

diversemix (approved):
Looks good

diversemix (approved):
lgtm :safety_vest:

PR #54, opened 2023-12-06
Feature/pdct 654 return empty 200 on all search entities when no results foun

katybaulch:
# Description

Please include:

- Extended search for collection, event, and document entities.
- Returned 200 instead of 404 when no matching entries found in DB.
- Added tests for all of this.
-  Linear ticket [PDCT-654](https://linear.app/climate-policy-radar/issue/PDCT-654/return-empty-200-on-all-search-entities-when-no-results-found)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

Added unit and integration tests.

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/collection.py
+++ b/app/api/api_v1/routers/collection.py
@@ -4,6 +4,11 @@ import logging
 from fastapi import APIRouter, HTTPException, Request, status
 
 import app.service.collection as collection_service
+from app.api.api_v1.query_params import (
+    get_query_params_as_dict,
+    set_default_query_params,
+    validate_query_params,
+)
 from app.errors import RepositoryError, ValidationError
 from app.model.collection import (
     CollectionCreateDTO,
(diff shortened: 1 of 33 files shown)


app/api/api_v1/routers/collection.py
```diff
@@ -69,27 +75,50 @@ async def get_all_collections() -> list[CollectionReadDTO]:
...
+    :raises HTTPException: If a DB error occurs a 503 is returned.
+    :raises HTTPException: If the search request times out a 408 is
+        returned.
+    :return list[CollectionReadDTO]: A list of matching collections
+        (which can be empty).
     """
+
+    query_params = get_query_params_as_dict(request.query_params)
+
+    DEFAULT_SEARCH_FIELDS = ["title", "description"]
```
diversemix, line 95: I wonder if you could explain `DEFAULT_SEARCH_FIELDS` for all the entities. It appears as though this is used to remove duplicates from the query_params - but it seems in the repository for the entities these fields are hard coded - so not sure why we use them?

katybaulch: Very good point. Updated.

PR #60, opened 2024-01-16
PDCT-683 only_allow_valid_url

asacristani:
# Description

Add Pydantic types for URL instead of string.
- Modify str for AnyHttpUrl
- Adapt the strings in tests (start with "http://").
- Include cast for match types during instantiations.
- Modify dumps mode to "json" instead of "python" default.
- Include casting pre database creation and update of a document (url pydantic types are not explicitly serialisable by sqlalchemy).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

Unit and integrity tests

## Reviewer Checklist

- [x] No FIXMEs remain


katybaulch (approved):
Looks good to me (LGTM) 🎋

app/repository/document.py
```diff
@@ -125,7 +125,7 @@ def _document_tuple_from_dto(db: Session, dto: DocumentCreateDTO) -> CreateObjec
     phys_doc = PhysicalDocument(
         id=None,
         title=dto.title,
-        source_url=dto.source_url,
+        source_url=str(dto.source_url),
```
diversemix, line 128: Here you are creating a new string based on the object (which itself is a string) - if you use `cast()` this is just generating a new memoryview. Overall this should have no effect, but its worth noting what you've written here could allocate more memory un-necessarily.

katybaulch: Oooo, this is really good to know!

diversemix (approved):
lgtm :)

app/repository/document.py
asacristani: I am confusing with cast() from typing.
I did the changes you mentioned (str() -> cast(str, ...)), but these are not working
I have read about better about cast, and basically cast is a function to make the linter trust in the type when these are not match, but do not change the type.
Therefore, the variables are being still URL from pydantic, and sqlalchemy is still not understanding them...
 
`sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) can't adapt type 'pydantic_core._pydantic_core.Url'`

It's important to note that cast() does not perform any runtime conversion operation

integration_tests/document/test_create.py
```diff
@@ -15,7 +15,7 @@ def test_create_document(client: TestClient, test_db: Session, user_header_token
     new_document = create_document_create_dto(title="Title", family_import_id="A.0.0.3")
     response = client.post(
         "/api/v1/documents",
-        json=new_document.model_dump(),
+        json=new_document.model_dump(mode="json"),
```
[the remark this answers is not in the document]
asacristani: model_dump return a dict and model_dump_json return a str.
model_dump have two methods, "python"(default) and "json":
- python: the dictionary may contain any Python objects
- json: the dictionary will only contain JSON serializable types

PR #61, opened 2024-01-16
PDCT-762 Changed default doc status back to CREATED.

katybaulch:
# Description

We jumped the gun changing the default doc status to PUBLISHED. As part of the single production truth work, we need the default to be CREATED.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

Updated integration test.

## Reviewer Checklist

- [x] No FIXMEs remain

### DIFF
--- a/integration_tests/document/test_create.py
+++ b/integration_tests/document/test_create.py
@@ -244,7 +244,7 @@ def test_create_document_when_invalid_variant(
     assert '(Invalid) is not present in table "variant"' in data["detail"]
 
 
-def test_document_status_is_published_on_create(
+def test_document_status_is_created_on_create(
     client: TestClient, test_db: Session, user_header_token
 ):
     setup_db(test_db)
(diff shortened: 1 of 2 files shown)


integration_tests/document/test_create.py
```diff
@@ -263,4 +263,4 @@ def test_document_status_is_published_on_create(
     )
 
     assert actual_fd is not None
-    assert actual_fd.document_status is DocumentStatus.PUBLISHED
+    assert actual_fd.document_status is DocumentStatus.CREATED
```
diversemix, line 266: Maybe we should change the test name too?

diversemix (approved):

integration_tests/document/test_create.py
katybaulch: Whoops - I was over zealous. Updated.

PR #62, opened 2024-01-24
PDCT-716/Make-library-for-alembic-db-schema

asacristani:
# Description

## DONE
1. Remove client/db folder
2. Start to use [navigator-db-client](https://github.com/climatepolicyradar/navigator-db-client/tree/v.0.0.1) library
3. Adapt code (including tests)
4. Move tests (no integrity) that affects only to the library -> NO TEST FOUND
5. Check the compatibility when navigator-backend will be adapted

## TO DISCUSS
- pip/poetry in Dockers -> Poetry
- RepositoryError -> Errors stay in backends

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change

## How Has This Been Tested?

Unit and integration.
Staging test before Production.

## Reviewer Checklist

- [x] No FIXMEs remain


app/api/api_v1/routers/analytics.py
```diff
@@ -1,7 +1,7 @@
 """Endpoints for managing the Analytics service."""
 import logging
 from fastapi import APIRouter, HTTPException, status
-from app.errors import RepositoryError
+from app.clients.db.errors import RepositoryError
```
diversemix, line 4: Hmm... I thought we were leaving the `RepositoryError` in here? Why did we move it again?

pyproject.toml
```diff
@@ -28,6 +28,7 @@ boto3 = "^1.28.46"
 moto = "^4.2.2"
 types-sqlalchemy = "^1.4.53.38"
 urllib3 = "^1.26.17"
+db-client = {git = "https://github.com/climatepolicyradar/navigator-db-client.git", tag = "main-bd82dba6"}
```
diversemix, line 31: Can we use a semver tag please :pray:

app/api/api_v1/routers/analytics.py
asacristani: Reverted!

.github/workflows/ci.yml
```diff
@@ -36,15 +36,15 @@ jobs:
         run: |
           poetry run pre-commit run --all-files
 
+      - name: Use .env.example
```
katybaulch, line 39: Not sure this does anything? We aren't telling the make build command to use the env file, so this is redundant

Dockerfile
```diff
@@ -1,13 +1,12 @@
 FROM python:3.9-slim
 
 WORKDIR /usr/src
-ENV PYTHONPATH=/usr/src
+ENV PYTHONPATH=/usr/src
```
katybaulch, line 4: Trailing whitespace

app/config.py
```diff
@@ -9,4 +9,8 @@
 SQLALCHEMY_DATABASE_URI = (
     f"postgresql://{_creds}@{ADMIN_POSTGRES_HOST}:5432/{ADMIN_POSTGRES_DATABASE}"
 )
+
+# This used by db-client and alembic script for migrations
+os.environ["DATABASE_URL"] = SQLALCHEMY_DATABASE_URI
```
katybaulch, line 14: Should we be setting this in the script, or in the environment we are running the admin service from?

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '3.7'
```
katybaulch, line 1: What's the rationale behind including a docker-compose file here now that we've moved to have a single source of truth by combining the admin service and backend databases?

app/initial_data.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
katybaulch, line 1: See comment on docker-compose

app/config.py
diversemix: I think the idea is to build all the necessary config from the `ADMIN_...` vars - unfortunately the only way of communicating this down to the `db-client` library seems to be setting the env var `DATABASE_URL` ... the `db-client` has a function `get_url()` which returns this value.

diversemix: So you can see from line 10 that the environment specifies the ADMIN vars and the code worries about formatting it up for the rest of the code. This make sense @katybaulch

docker-compose.yml
diversemix: I guess to fully de-couple the two things and also letting them develop independently. 

For example: If the nav-backend breaks the ability to spin up a local db - then admin development can still continue. Arguably - we could move this into the `db-client` in some way in the future. But for now I think this is the simplest/less risky approach.

Happy to discuss further.

app/initial_data.py
diversemix: See https://github.com/climatepolicyradar/navigator-admin-backend/pull/62#discussion_r1489345597 :smile:

diversemix (approved):
I'm happy - as long as we get another review :grin:

app/config.py
katybaulch: Ah understood, thanks for clarifying

katybaulch (approved):
All good, well done! 🦖

PR #74, opened 2024-02-21
Fix auto semver tagging

katybaulch:
# Description

- Created new workflow 'tag' with two jobs, one to run on manually created tags & one to run on auto-generated tags
- Removed existing semver and main-push-autotag workflows
- Resolved workflow dependencies

## Proposed version

- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] No FIXMEs remain


.github/workflows/tag.yml
```diff
@@ -46,8 +48,37 @@ jobs:
...
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: eu-west-1
+
+    - name: Login to Amazon ECR
+      id: login-ecr
+      uses: aws-actions/amazon-ecr-login@v1.6.1
```
olaughter, line 60: Theres a v2 release with some security updates that might be worth trying out

.github/workflows/tag.yml
```diff
@@ -46,8 +48,37 @@ jobs:
...
+    - name: Docker pull, retag and push
+      env:
+        NEW_TAG: ${{ steps.run_git_auto_tag.outputs.new_tag }}
+        DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      run: |
+        docker pull ${DOCKER_REGISTRY}/navigator-admin-backend:main-${GITHUB_SHA::8}
+        docker tag ${DOCKER_REGISTRY}/navigator-admin-backend:main-${GITHUB_SHA::8}  navigator-admin-backend:main-${GITHUB_SHA::8}
+        .github/retag-and-push.sh navigator-admin-backend main-${GITHUB_SHA::8}
```
olaughter, line 71: This is a neat way of getting the shortened sha

.github/workflows/tag.yml
```diff
@@ -46,8 +48,37 @@ jobs:
...
+
+    - name: Create and push Git tag
+      run: |
+        new_tag=${{ steps.run_git_auto_tag.outputs.new_tag }}
+        new_version_num=${new_tag#v}
+        git tag -a "${new_tag}" -m "Version ${new_version_num}"
+        git push --tags origin "${new_tag}"
+
+  semver-for-manually-created-tags:
+    if: ${{ startsWith(github.ref, 'refs/tags/') && github.event.workflow_run.conclusion == 'success' }}
```
olaughter, line 81: I'm not sure I understand this line, are we checking if it is a release that ran successfully? If so I'm not sure how `github.event.workflow_run.conclusion` will behave, how does it know which workflow_run to target? Maybe this is not a problem (Or I've misunderstood!), but another approach would be to run as a dependant job using the `needs` key: https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs

.github/workflows/tag.yml
```diff
@@ -1,13 +1,15 @@
...
+name: Auto-Tag
 
 on:
-  push:
-    branches:
-      - main
+  workflow_run:
+    workflows: [build]
+    types:
+      - completed
```
olaughter, line 7: Ahha I think my question in my previous comment just clicked that the event is in scope because of the specific way this is being triggered. Feel free to disregard

olaughter (approved):
Great, lets give it a whirl!

.github/workflows/tag.yml
katybaulch: Great to know - let me whack that in now!

.github/workflows/tag.yml
katybaulch: It's currently not possible to use needs when the job that you want your job to depend on lives in a different workflow 😢 This new tag workflow runs after the 'build' job in the CI workflow has executed - but only if certain other conditions are met. The 'on' at the top of a workflow file invokes the workflow for events joined by a logical OR, not AND, so that's why we're using the if config option here. The workflow_run.conclusion context variable therefore stores whether the 'build' job from the CI workflow completed successfully or failed. We need this because 'completed' just means its finished running rather than distinguishing what the result is.

PR #75, opened 2024-02-21
PDCT-853 Fix admin backend tests & don't ignore test errors in CI

katybaulch:
# Description

- Fixed error reporting in CI by removing '-' from beginning of makefile commands
- Fixed tests

## Proposed version

- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [x] No FIXMEs remain


Makefile
```diff
@@ -19,24 +19,24 @@ test_bashscripts: build_bats
...
 run: 
-	- docker-compose -f docker-compose.yml up -d --remove-orphans
+	docker-compose -f docker-compose.yml up -d --remove-orphans
 
 start: build run migrations
 
 show_logs: 
-	- docker-compose logs -f
\ No newline at end of file
+	docker-compose logs -f
```
diversemix, line 42: We probably want to keep the hyphen on this - as showing the logs is not a gating action

app/repository/document.py
```diff
@@ -125,7 +125,7 @@ def _document_tuple_from_dto(db: Session, dto: DocumentCreateDTO) -> CreateObjec
     phys_doc = PhysicalDocument(
         id=None,
         title=dto.title,
-        source_url=cast(str, dto.source_url) if dto.source_url is not None else None,
+        source_url=str(dto.source_url) if dto.source_url is not None else None,
```
diversemix, line 128: Can we mark a TODO on this ... and that is to do some more verification here - unless you want to do that now ... it would involve ensuring that if its not None - then its a valid url

app/repository/family.py
```diff
@@ -79,7 +79,7 @@ def _family_to_dto(db: Session, fam_geo_meta_org: FamilyGeoMetaOrg) -> FamilyRea
         category=str(f.family_category),
         status=str(f.family_status),
         metadata=metadata,
-        slug=str(f.slugs[0].name if len(f.slugs) > 0 else ""),
+        slug=str(f.slugs[-1].name if len(f.slugs) > 0 else ""),
```
diversemix, line 82: Let's check why we need this change

Makefile
katybaulch: Okie dokie, pushing now:)

app/repository/document.py
katybaulch: [PDCT-865](https://linear.app/climate-policy-radar/issue/PDCT-865/add-more-verification-for-source-url)

app/repository/family.py
katybaulch: reverted

diversemix (approved):
great

PR #76, opened 2024-02-21
Fix auto semver tagging

katybaulch:
# Description

- Corrected workflow name to 'CI' instead of build
- Bumped setup_python version to 5 

## Proposed version

- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Reviewer Checklist

- [ ] No FIXMEs remain


.github/workflows/tag.yml
```diff
@@ -2,10 +2,13 @@ name: Auto-Tag
 
 on:
   workflow_run:
-    workflows: [build]
+    workflows: ["CI"]
     types:
       - completed
 
+  push:
+    tags: ['v[0-9]+.[0-9]+.[0-9]+']
```
olaughter, line 10: Will this catch `beta` suffixes, like:`v1.1.1-beta`

olaughter (approved):

.github/workflows/tag.yml
katybaulch: Oooh good point. Reverting this to v* as I'm not confident enough in the pattern above now..

PR #107, opened 2024-03-26
PDCT 998/ensure new documents have relationships in old and new schema

diversemix:
# Description

- Upgrade dbclient to v3.2.1 & remove ruff config
- Configure pyright to work with pyenv in makefile
- Add pyrightconfig to gitignore & add trunk ignores
- Add ruff config for trunk
- Populate old and new schema for new families.
- Don't add corpus type on family create.
- Update blank.sql with alembic version 0032 schema
- Fix configure pyright & pyenv script
- Use new function scoped db fixture that runs the db migrations ⚠️ - this is 3x slower than before :(

Co-authored-by: @diversemix 

## Proposed version

- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


.trunk/configure-pyright-with-pyenv.sh
```diff
@@ -7,22 +7,19 @@ venv_name=$(grep -m 1 venv pyproject.toml | tr -s ' ' | tr -d '"' | tr -d "'" |
...
 		echo "pyenv not installed. Please install pyenv..."
 		exit 1
 	fi
 
-	# Check if pyenv-pyright plugin is installed
-	if ! command -v pyenv pyright &>/dev/null; then
-		echo "pyenv-pyright not installed. Installing..."
-		pyenv_root=$(pyenv root)
+	pyenv_root=$(pyenv root)
+	if [ -z $(ls -A "${pyenv_root}"/plugins/pyenv-pyright) ]; then
```
diversemix, line 16: If you type `man test` you'll find:
```
       -d FILE
              FILE exists and is a directory
```
which may be a little neater

integration_tests/family/test_create.py
```diff
@@ -24,28 +25,47 @@ def test_create_family(client: TestClient, data_db: Session, user_header_token):
...
-    import_id = response.json()
-    assert import_id == "CCLW.family.i00000001.n0000"
-    actual_family = data_db.query(Family).filter(Family.import_id == import_id).one()
+    expected_import_id = "CCLW.family.i00000001.n0000"
+    assert response.json() == expected_import_id
+    actual_family = (
+        data_db.query(Family).filter(Family.import_id == expected_import_id).one()
+    )
 
     assert actual_family.title == "Title"
```
diversemix, line 34: Do we also want to test the `actual_family.organisation_id` here too - just to ensure its still being used'?

.trunk/configure-pyright-with-pyenv.sh
katybaulch: I specifically want to test that the dir exists **and** that the dir isn't empty. Not sure I can do that with just `test`?

integration_tests/family/test_create.py
katybaulch: Good shout

diversemix:
All looking good from my side of the :person_fencing:

katybaulch (approved):
Peter raised, so I need to close it off 😆 Lovely work from us both if I do say so 😉

PR #129, opened 2024-05-02
Feature/pdct 1055 frontend create family enable the user to select the corpus

katybaulch:
# Description

Enable user to select corpus on create in the API.

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

Existing tests updated, new tests added.

## Reviewer Checklist

- [ ] No FIXMEs remain


app/model/family.py
```diff
@@ -23,7 +23,7 @@ class FamilyReadDTO(BaseModel):
     documents: list[str]
     collections: list[str]
     organisation: str
-    corpus_id: str
+    corpus_import_id: str
```
diversemix, line 26: Good shout - I did call this `corpus_id` in teh nav frontend as it doesn't use `import_id` for anything

diversemix (approved):
Looks great - I know that getting `call_count` right in those tests can be a bit of a pain :medal_military:

app/model/family.py
katybaulch: Nice! I contemplated but I think this is more consistent with how we're doing it in this repo, so thought I'd change for now!

PR #147, opened 2024-06-05
Pdct 1066 user org not checked on document delete

katybaulch:
# Description

Check user org matches org the document belongs to on document delete.

## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/repository/document.py
+++ b/app/repository/document.py
@@ -480,3 +480,12 @@ def count(db: Session, org_id: Optional[int]) -> Optional[int]:
         return
 
     return n_documents
+
+
+def get_org_from_import_id(db: Session, import_id: str) -> Optional[int]:
+    result = _get_query(db).filter(FamilyDocument.import_id == import_id).one_or_none()
+    _LOGGER.error(result)
+    if result is None:
+        return None
+    _, _, org, _, _ = result
+    return org.id
--- a/app/service/document.py
+++ b/app/service/document.py
@@ -150,7 +150,7 @@ def create(
 @validate_call(config=ConfigDict(arbitrary_types_allowed=True))
 def delete(
     import_id: str, user_email: str, context=None, db: Session = db_session.get_db()
-) -> bool:
+) -> Optional[bool]:
     """
     Deletes the document specified by the import_id.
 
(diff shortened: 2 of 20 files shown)


app/repository/document.py
```diff
@@ -480,3 +480,12 @@ def count(db: Session, org_id: Optional[int]) -> Optional[int]:
...
 
     return n_documents
+
+
+def get_org_from_import_id(db: Session, import_id: str) -> Optional[int]:
+    result = _get_query(db).filter(FamilyDocument.import_id == import_id).one_or_none()
+    _LOGGER.error(result)
+    if result is None:
+        return None
+    _, _, org, _, _ = result
```
diversemix, line 490: I never knew you could use `_` multiple times ... noice!

app/service/document.py
```diff
@@ -148,16 +148,36 @@ def create(
...
+
+    doc = get(import_id)
+    if doc is None:
+        return None
+
     if context is not None:
         context.error = f"Could not delete document {import_id}"
+
+    entity_org_id = get_org_from_id(db, import_id)
+    app_user.is_authorised_to_make_changes(db, user_email, entity_org_id, import_id)
```
diversemix, line 173: Just a note - I wonder if the return value of this function should be removed - and the function renames to something like `raise_if_not_in_organisation` ... just that the function as-is doesn't actually return false so could be misleading - not a blocker - food for thought

diversemix (approved):
lgtm + :speech_balloon:

app/service/document.py
katybaulch: Yeah I like that idea - I might implement in the latest PR to avoid a bunch of merge conflicts though :)

PR #153, opened 2024-06-10
Add user's org to the token

diversemix:
# Description

- add org to token
- update tests
- fix is_admin to always be True as per real life

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/service/authentication.py
```diff
@@ -59,6 +60,11 @@ def authenticate_user(email: str, password: str) -> str:
             _LOGGER.error(f"Failed login attempt as inactive {email}")
             raise AuthenticationError(f"User {email} is marked as not active.")
 
+        org_id = app_user_repo.get_org_id(db, cast(str, user.email))
+        if org_id is None:
+            _LOGGER.error(f"Failed login attempt, org not found {email}")
```
THOR300, line 65: `...not found for {email}` might read better

tests/integration_tests/conftest.py
```diff
@@ -218,34 +220,34 @@ def rollback_event_repo(monkeypatch, mocker):
...
 def superuser_header_token() -> Dict[str, str]:
-    a_token = token_service.encode("super@cpr.org", True, {})
+    a_token = token_service.encode("super@cpr.org", ORG_ID, True, {})
     headers = {"Authorization": f"Bearer {a_token}"}
     return headers
 
 
 @pytest.fixture
 def user_header_token() -> Dict[str, str]:
-    a_token = token_service.encode("cclw@cpr.org", False, {"is_admin": False})
```
THOR300, line 228: I'm interested why this `org_id` update now means all users are admins?

THOR300 (dismissed):
Non blocking query about why all users are admins now! 

And great that the org id is just an integer and thus non-sensitive data! 

Is it a concern that we are encoding email addresses?

katybaulch:
> Non blocking query about why all users are admins now!
> 
> And great that the org id is just an integer and thus non-sensitive data!
> 
> Is it a concern that we are encoding email addresses?

Hey, it's two separate changes! So we want to add the org_id to the user token so we can access it easily in the backend. The driveby fix here is to make bump the test users to admin to reflect the real life database. Admin users are able to use the admin service but are restricted to CRUDding entities belonging to the same organisation as themselves, whereas superuser can CRUD all entities across all organisations.

Not at all luckily, as it's all in the JWT token which can't be decoded without the secret key!

THOR300:
> > Non blocking query about why all users are admins now!
> > And great that the org id is just an integer and thus non-sensitive data!
> > Is it a concern that we are encoding email addresses?
> 
> Hey, it's two separate changes! So we want to add the org_id to the user token so we can access it easily in the backend. The driveby fix here is to make bump the test users to admin to reflect the real life database. Admin users are able to use the admin service but are restricted to CRUDding entities belonging to the same organisation as themselves, whereas superuser can CRUD all entities across all organisations.
> 
> Not at all luckily, as it's all in the JWT token which can't be decoded without the secret key!

I only ask as I was doing some quick reading around as I've not personally used JWT's in this much detail before, always been a consumer of them rather than a creator! 

"When it comes to access tokens, they can easily be replaced by opaque tokens (as mentioned in the previous section), but the ID token is always a JWT. This means that you should put extra care into what is available in the JWT so that no sensitive data is unintentionally revealed. It will be much safer for your UI client to call the user info endpoint and get the user's data from there instead of keeping it directly in the ID token." 

My understanding of this was that if we want the user data we can just get that after the frontend receives the token? 

https://curity.io/resources/learn/jwt-best-practices/

diversemix:
@THOR300 ... you are of course right and this statement (which I said to Katy in the middle of coding) is wrong: "Not at all luckily, as it's all in the JWT token which can't be decoded without the secret key!" ... what I should have said is that its impossible for the user to fake a token - i.e. the user or anyone who has the token can decode it ... but they cannot fake a token without having the private key... so that for instance they cannot set themselves to belong to another organisation. 

Second point - we could use the user's db id to identify them rather than their email - but we don't use "real" emails  anyway. This is a good point if we should do - so gonna raise a ticket - thanks.

diversemix:
See https://linear.app/climate-policy-radar/issue/PDCT-1174/when-using-real-emails-do-not-use-in-the-jwt

THOR300 (approved):
LGTM

PR #156, opened 2024-06-11
Feature/pdct 1161 user org not checked on family update

katybaulch:
# Description

- Validate user org on family update

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/tests/integration_tests/family/test_update.py
+++ b/tests/integration_tests/family/test_update.py
@@ -249,7 +249,7 @@ def test_update_family_collections_to_one_that_does_not_exist(
     assert db_collections[0].collection_import_id == "C.0.0.2"
 
 
-def test_update_family_when_user_org_different_to_family_org(
+def test_update_fails_family_when_user_org_different_to_family_org(
     client: TestClient, data_db: Session, non_cclw_user_header_token
 ):
     setup_db(data_db)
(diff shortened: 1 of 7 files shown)


tests/integration_tests/family/test_update.py
```diff
@@ -282,6 +282,51 @@ def test_update_family_when_user_org_different_to_family_org(
...
+    new_family = create_family_write_dto(
+        title="apple",
+        summary="just a test",
+        geography="USA",
+        category=FamilyCategory.UNFCCC,
+        metadata={"color": ["pink"], "size": [0]},
+        collections=["C.0.0.3"],
+    )
+    response = client.put(
+        "/api/v1/families/A.0.0.1",
```
THOR300, line 298: This line jumped out to me. For my own understanding is it good practice to have `.` in an api call? 
Would it be better to add the collection in future to the json data or something?

THOR300 (approved):
Unreal! Non blocking comment

tests/integration_tests/family/test_update.py
katybaulch: > "[Characters that are allowed in a URI but do not have a reserved purpose are called unreserved. These include uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde](https://www.rfc-editor.org/rfc/rfc3986)"

Full stops are allowed generally, but I think the situations to avoid are instances where a period or double period are sandwiched in between a part of the URI on their own e.g., `www.cpr.com/family/./` and `www.cpr.com/family/../` would redirect to `www.cpr.com/family/` and `www.cpr.com/` respectively.

> Would it be better to add the collection in future to the json data or something?

The collection "C.0.0.3" is already being added to the family "A.0.0.1" via JSON (line 299 in this case `json=new_family.model_dump()`) :)

diversemix (approved):
re-approve

PR #171, opened 2024-07-08
Feature/pdct 1139 ensure document metadata is validated in db client

katybaulch:
# Description

- Validate document metadata using db_client
- Driveby validate family metadata using db_client
- Updated tests 

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/repository/document.py
```diff
@@ -133,6 +134,11 @@ def _doc_to_dto(doc_query_return: ReadObj) -> DocumentReadDTO:
         created=cast(datetime, fdoc.created),
         last_modified=cast(datetime, fdoc.last_modified),
         slug=str(fdoc.slugs[0].name if len(fdoc.slugs) > 0 else ""),
+        metadata=(
+            cast(dict, fdoc.valid_metadata)
+            if fdoc.valid_metadata is not None
+            else {"role": []}
```
diversemix, line 140: I wonder if we should pull out a `default_document_metadata` variable for this?

diversemix: Another thought - why do we need "role" on there? Why can't it be an empty dict? Or is this just temp while we change over?

app/service/document.py
```diff
@@ -124,6 +126,18 @@ def update(
...
+    family = family_service.get(doc.family_import_id)
+    if family is None:
+        raise ValidationError(f"Could not find family for {doc.family_import_id}")
+
+    # Validate metadata.
+    metadata_service.validate_metadata(
+        db,
+        family.corpus_import_id,
+        document.metadata,
+        EntitySpecificTaxonomyKeys.DOCUMENT.value,
```
diversemix, line 138: Oh no - was the enum not based on str (see SortOrder)

app/repository/document.py
katybaulch: Can do. Worth noting that my next PR after I've tested this on staging will be to do an UPDATE on the documents table to create a JSON agg object to populate the metadata field if it's NULL by reading from the value in the role column  - so might be wasted effort

app/service/metadata.py
```diff
@@ -0,0 +1,38 @@
...
+    :param str corpus_id: The corpus ID to get the taxonomy for.
+    :param TaxonomyDataEntry metadata: The metadata to validate.
+    :param Optional[str] entity_key: The entity specific taxonomy key if
+        exists, otherwise None (which represents validating Family
+        metadata).
+    :raises ValidationError: if the metadata is invalid.
+    :return None
+    """
+    try:
+        results = metadata_repo.validate_metadata(db, corpus_id, metadata, entity_key)
```
diversemix, line 31: Is it appropriate to still call this `metadata_repo` ? ... not sure ... resolve if you are happy!

tests/unit_tests/service/family/test_update_family_service.py
```diff
@@ -373,7 +411,8 @@ def test_update_success_when_family_organisation_mismatch_with_user_org(
     assert geography_repo_mock.get_id_from_value.call_count == 1
     assert family_repo_mock.get.call_count == 2
     assert corpus_repo_mock.get_corpus_org_id.call_count == 1
-    assert corpus_repo_mock.get_taxonomy_from_corpus.call_count == 1
+    assert metadata_repo_mock.get_taxonomy_from_corpus.call_count == 1
```
diversemix, line 414: Job for another time - but not convinced that this tests needs to confirm some of these calls :thinking:

diversemix (approved):
I'm happy - one or two questions tho

app/service/document.py
katybaulch: Yes it is - The function `get_entity_specific_taxonomy` takes a str though not the enum, so I've had to do this until I can update  it in the db_client as a driveby

tests/unit_tests/service/family/test_update_family_service.py
katybaulch: Think you're right - they were originally to help debugging the tests with the mocks. Lets review as part of wider test strat discussion next week?

app/service/metadata.py
katybaulch: Happy to change

app/repository/document.py
diversemix: yeah happy for you to leave as-is if that is the case

app/service/document.py
diversemix: thanks

tests/unit_tests/service/family/test_update_family_service.py
diversemix: sgtm

app/service/metadata.py
diversemix: Let's leave it as is ... I don't think there is a strong argument either way - so let's not fix what isn't broken :smile:

diversemix (approved):
Nice :+1:

diversemix (approved):
great

PR #254, opened 2024-11-25
Feature/pdct 1570 add aws s3 docker image for integration tests

annaCPR:
# Description
- add localstack to create a local version of AWS S3 service for integration testing
- add a fixture to cleanup any files created and saved to the local S3 bucket during tests
- add a pytest annotation to plug in the S3 cleanup fixture to integration tests
- remove old logic and related test for skipping S3 backup in local development
- update README

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


init-s3.py
katybaulch, line 1: Can this go in the tests folder, or even in the integration tests folder?

GETTING_STARTED.md
```diff
@@ -72,14 +77,15 @@ how the tests run in the Github Actions CI pipeline.
 
 ### Integration Tests
 
-These tests are designed to require a database and therefore will pull and run a
-Postgres container. These can be run locally with
-`pytest -vvv integration_tests` - however this will require that you have spun
-up a local postgres instance (which can be done using `make setup_test_db`).
+These tests are designed to require a database as well as an AWS S3 mock service
+providedvia localstack and therefore will pull and run Postgres and localstack
```
katybaulch, line 81: Typo
```suggestion
provided via localstack and therefore will pull and run Postgres and localstack
```

GETTING_STARTED.md
```diff
@@ -19,6 +19,11 @@ Create a `.env` file, you can use the example one running:
 cp .env.example .env
 ```
 
+**NOTE**: When running the app in docker, different values are required for some
+environment variables (as per comments in .env.example). So, if running the app
```
katybaulch, line 23: Are we ever running the app or tests without Docker?

init-s3.py
```diff
@@ -0,0 +1,11 @@
+import boto3
+
+# trunk-ignore(bandit/B106)
+s3_client = boto3.client(
+    "s3",
+    endpoint_url="http://localhost:4566",
+    aws_access_key_id="test",
```
katybaulch, line 7: Same as above.

init-s3.py
```diff
@@ -0,0 +1,11 @@
...
+    aws_secret_access_key="test",
```
katybaulch, line 8: Same as above.

init-s3.py
```diff
@@ -0,0 +1,11 @@
...
+)
+
+s3_client.create_bucket(Bucket="bulk-import-bucket")
```
katybaulch, line 11: Same as above.

pyproject.toml
```diff
@@ -54,7 +54,7 @@ env_files = """
     .env.test
     .env
 """
-markers = ["unit"]
+markers = ["unit", "integration"]
```
katybaulch, line 57: Instead of marking all these tests as integration tests using pytest, can we mark them as S3 tests instead and be more explicit? To run all the integrations tests, we would run pytest on the whole folder -- so this 'mark' isn't currently representative of what it says it's going to do

tests/unit_tests/conftest.py
```diff
@@ -290,37 +292,59 @@ def test_s3_client(s3_document_bucket_names):
...
             )
-        yield conn
+            try:
+                conn.head_bucket(Bucket=bucket_name)
+            except ClientError:
+                conn.create_bucket(
+                    Bucket=bucket_name,
+                    CreateBucketConfiguration={"LocationConstraint": "eu-west-2"},
+                )
+                conn = boto3.client("s3", region_name="eu-west-2")
```
katybaulch, line 346: Can the region be a constant (or in the .env file if it's used in the source code)

init-s3.py
annaCPR: Good spot

init-s3.py
annaCPR: Good spot

init-s3.py
annaCPR: Good spot

GETTING_STARTED.md
```diff
@@ -72,14 +77,15 @@ how the tests run in the Github Actions CI pipeline.
...
 ### Integration Tests
 
-These tests are designed to require a database and therefore will pull and run a
-Postgres container. These can be run locally with
-`pytest -vvv integration_tests` - however this will require that you have spun
-up a local postgres instance (which can be done using `make setup_test_db`).
+These tests are designed to require a database as well as an AWS S3 mock service
+providedvia localstack and therefore will pull and run Postgres and localstack
+containers.
+These can be run locally with `pytest -vvv integration_tests` - however this will
```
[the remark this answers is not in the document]
annaCPR: I've actually been running my integration tests this way until now and we do say what the recommended way to run them is (i.e. make integration_tests), and why, in this file. Are we sure it is no longer needed and no one will want to set themselves up to run them with pytest directly? If so, I'll remove it.

GETTING_STARTED.md
annaCPR: See comment above

.env.example
```diff
@@ -1,19 +1,20 @@
...
 SECRET_KEY=secret_test_key
 API_HOST=http://navigator-admin-backend:8888
 
 # AWS
-BULK_IMPORT_BUCKET=skip
+BULK_IMPORT_BUCKET=bulk-import-bucket
 AWS_ACCESS_KEY_ID=test
-_AWS_SECRET_ACCESS_KEY=test
+AWS_SECRET_ACCESS_KEY=test
+AWS_ENDPOINT_URL=http://localstack:4566 # for running tests outside docker this needs to be http://localhost:4566
```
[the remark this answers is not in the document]
annaCPR: It is the default value for local dev. Not sure we'd ever need/want to connect our local app to actual AWS in staging or prod, can't think of a use case...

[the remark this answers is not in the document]
katybaulch: Then can we update this comment to say 'Do not change this value, even if running outside of Docker?' or 'This value should be the same even if you're running your tests outside of Docker'?

.env.example
```diff
@@ -1,19 +1,20 @@
...
 PGUSER=navigator
 PGPASSWORD=password
-PGHOST=admin_backend_db
+PGHOST=admin_backend_db # for running tests outside docker this needs to be localhost
 
 # API
 SECRET_KEY=secret_test_key
 API_HOST=http://navigator-admin-backend:8888
 
 # AWS
```
katybaulch, line 13: We could update this comment to make it more explicit/descriptive about what these are used for? Are all of these only used for local development/testing? Or are some used by the actual application?

katybaulch: As someone who has not worked on bulk import, it is unclear what these are for and where/when they are used.

GETTING_STARTED.md
katybaulch: Not sure that answers my question I'm afraid - my question wasn't constrained by AWS. Do we ever run the admin backend application outside of Docker e.g., just by running `python app/main.py` in our terminal? Or is the development flow that we always use Docker for running our application?

GETTING_STARTED.md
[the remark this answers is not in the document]
katybaulch: I think we need to have a team conversation about this - let's leave it for now and chat next week :) /cc @jamesgorrie as this is something we've spoken about briefly

.env.example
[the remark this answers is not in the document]
annaCPR: We might be misunderstanding each other here. If we want to still support running of tests both inside and outside of docker then the value will be different in each case (unless you know how I can make it the same). Inside docker the expected endpoint is localstack:4566 but outside docker it should be localhost:4566. However, if you think that no one will need to run the tests outside of docker and the additional set up is not necessary then I can remove this comment.

katybaulch (approved):
lgtm - questions around env vars but otherwise all good!

PR #255, opened 2024-11-26
Add pgAdmin to docker compose for nice local admin UI

jamesgorrie:
# Description

- adds [pgAdmin](https://www.pgadmin.org/) to docker compose to give us a nice UI into the DB
  Screenshot 2024-11-26 at 17 38 36
- couple of [yamllint](https://yamllint.readthedocs.io/en/stable/) fixes
  
## Proposed version

- [x] Skip auto-tagging
- [ ] Patch
- [ ] Minor version
- [ ] Major version

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

This is a local change so can be tested by running `make run` and visiting http://localhost:8000

I'll be making some follow up changes, so I'll add some docs while there.

## Reviewer Checklist

- [ ] No FIXMEs remain


docker-compose.yml
```diff
@@ -32,6 +31,49 @@ services:
...
+    environment:
+      # This deavctivates the login screen
+      PGADMIN_CONFIG_SERVER_MODE: "False"
+      PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED: "False"
+    volumes:
+      - pgadmin-data:/var/lib/pgadmin
+    # This is just nice to know as if it fails, it fails quietly
+    healthcheck:
+      test: [CMD, wget, -O, "-", http://localhost:80/misc/ping]
+configs:
```
katybaulch, line 58: I'd love some more info on this `configs` section as I haven't come across it before! Wha's the benefit of this over extending the out of the box pgadmin4 image and configuring inside that extended image?

docker-compose.yml
```diff
@@ -32,6 +31,49 @@ services:
...
+  pgpass:
+    content: "admin_backend_db:5432:*:${POSTGRES_USER}:${POSTGRES_PASSWORD}"
+  servers.json:
+    content: |
+      {
+        "Servers": {
+          "1": {
+            "Group": "Servers",
+            "Name": "Navigator Admin Backend",
+            "Host": "admin_backend_db",
```
katybaulch, line 68: Should we be using `PGHOST` for this? What if someone changes the value in the .env file?

docker-compose.yml
```diff
@@ -32,6 +31,49 @@ services:
...
+      PGADMIN_CONFIG_SERVER_MODE: "False"
+      PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED: "False"
+    volumes:
+      - pgadmin-data:/var/lib/pgadmin
+    # This is just nice to know as if it fails, it fails quietly
+    healthcheck:
+      test: [CMD, wget, -O, "-", http://localhost:80/misc/ping]
+configs:
+  pgpass:
+    content: "admin_backend_db:5432:*:${POSTGRES_USER}:${POSTGRES_PASSWORD}"
```
katybaulch, line 60: Should we be using `PGHOST` for this? What if someone changes the value in the .env file?

docker-compose.yml
jamesgorrie: When you say extending the pg4admin image - what do you mean? Creating a new `Containerfile` for it?

My thinking on this was that pgAdmin makes these options configurable via files that exist on the filesystem at startup. A way that's recommended by Docker is to use [configs](https://docs.docker.com/reference/compose-file/services/#configs) ([here's the top level element](https://docs.docker.com/reference/compose-file/configs/)).

I was you could specify it directly on the specific `service` - but their API isn't shaped that way.

Not sure I've answered the question?

docker-compose.yml
jamesgorrie: I thought hardcoding this to what's in the `Dockerfile` made sense as `localhost` wouldn't work here as the networking is happening between the services specified in this file.

docker-compose.yml
jamesgorrie: I'll keep the discussion running in the above comment.

docker-compose.yml
katybaulch: Thanks for sharing - I hadn't come across this before! Good to know :)

katybaulch (changes requested):
Code quality checks are failing but otherwise happy -- following up from our chat just now in the office I think we might benefit from starting to write our own dockerfiles and store them in ECR -- as I imagine this might be reusable between repos :) Nice work!

katybaulch (approved):
lgtm, we probably want to fix the conflicting lint rules going forward though rather than ignoring

docker-compose.yml
```diff
@@ -46,5 +46,56 @@ services:
...
+      PGADMIN_CONFIG_SERVER_MODE: "False"
+      PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED: "False"
+    volumes:
+      - pgadmin-data:/var/lib/pgadmin
+    # This is just nice to know as if it fails, it fails quietly
+    healthcheck:
+      # there are conflicting trunk errors here. Either
+      # - having `:` in the URL (fails in Github actions)
+      # - redundantly quoting the URL (fails locally)
+      # trunk-ignore(yamllint/quoted-strings)
```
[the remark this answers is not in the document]
jamesgorrie: Yea, I would have tried a bit more, but it's blocking this branch which was getting conflicts.

The annoying thing about this conflict is that everything seems to pass locally, but I might need to try understand the different lifecycles of trunk, which is where my knowledge gap is.

PR #256, opened 2024-11-28
Geography filter on admin service does not work

jamesgorrie:
I'll be using this branch to break off smaller branches, and eventually close this off.

PR #257, opened 2024-11-28
fix geography filter on family search

jamesgorrie:
# Description
- uses a more standard join for Geography to fetch geographies on a family

The subquery pattern that was previously used was conflicting with the filter attached further down the file

Whilst doing this, I think I've got [firm footing](https://github.com/climatepolicyradar/navigator-admin-backend/pull/256) to get the rest of the muli-geo work in pretty quick.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

## How Has This Been Tested?

- [x] integration test
- [x] locally
- [x] on staging

https://github.com/user-attachments/assets/812ebe34-47f2-4c8b-a5d5-5123a55e2f9b


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "admin_backend"
-version = "2.17.16"
+version = "2.17.17"
```
katybaulch, line 3: This is out of date now :(

tests/integration_tests/family/test_search.py
```diff
@@ -7,6 +7,36 @@
...
+        headers=superuser_header_token,
+    )
+    assert response.status_code == status.HTTP_200_OK
+    data = response.json()
+    assert isinstance(data, list)
+
+    ids_found = set([f["import_id"] for f in data])
+    assert len(ids_found) == 2
+
+    # search for where we know there is 1 families with a geopraphy
```
katybaulch, line 27: Can we parametrize this test by passing in the geography and the expected ids found instead of doing two checks within the same test, as the test code is the same for both `zimbabwe` and `afghanistan`?

tests/integration_tests/family/test_update.py
```diff
@@ -279,7 +279,11 @@ def test_update_fails_family_when_user_org_different_to_family_org(
     assert db_family.description == "apple"
     assert db_family.family_category == "UNFCCC"
 
-    geo_id = data_db.query(Geography.id).filter(Geography.value == "Other").scalar()
+    geo_id = (
+        data_db.query(Geography.id)
+        # TODO: PDCT-1406: Properly implement multi-geography support
```
katybaulch, line 284: Nice link to the linear ticket!

katybaulch (changes requested):
Nice one! Would like the test I've commented on to be parametrized if possible - as the test code is essentially the same for both checks you're doing on Afghanistan and Zim :)

pyproject.toml
katybaulch: Not that that matters actually, because auto tagging only happens when you merge into main, and this is merging back into a branch that doesn't match that pattern 💡

tests/integration_tests/family/test_search.py
jamesgorrie: How do you feel about moving it to a table based test, which eeps the tests logic dry, but also avoids having to jump through an abstration to see what the test is doing?

tests/integration_tests/family/test_update.py
jamesgorrie: Stole it from around the code.

katybaulch (approved):
Looks good to me!!

jamesgorrie:
Fixes [PDCT-1445 Geography filter on admin service does not work](https://linear.app/climate-policy-radar/issue/PDCT-1445/geography-filter-on-admin-service-does-not-work)

PR #266, opened 2024-12-09
Add multi geographies to search API

jamesgorrie:
# Description

Part of https://linear.app/climate-policy-radar/issue/PDCT-1404/admin-service-support-multiple-geographies

Allows the API to receive geography as an list i.e. `geography=zambia&geography=zimbabwe`.

I've used the [built in parsers for the geography qs](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#query-parameter-list-multiple-values) - which breaks convention, but avoids us having to add new parsers to `get_query_params_as_dict`. 

As a secondary refactor to this PR - we could lean heavier into that and be more explicit about out URL params.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

## Type of change

- [x] New feature

## How Has This Been Tested?

Integration tests


jamesgorrie (commented):
A couple notes to hopefully make the review easier.

katybaulch (approved):
This adds the ability to search multiple geographies via the API, but doesn't add support for multi geographies in the admin backend (slight misinterpretation of the ticket perhaps - but this work still needed doing so I'm going to link the right ticket and merge this) :)

https://linear.app/climate-policy-radar/issue/PDCT-1775/make-geography-filter-on-family-list-page-support-multi-geos

PR #268, opened 2024-12-19
feature/pdct-1777-support-multiple-geos-in-get

odrakes-cpr:
# Description

Support multiple geographies in the get request. 

- Create a subquery that aggregates all the family geographies into a list, insert this subquery into the main query. 
- Update the family_dto function to extract those values
- To ensure backwards compatibility with the front-end add new property key with geographies, and for the current geography property select the first element of the list.
- Update tests accordingly 
- Duplicated the query for the search endpoint, whilst we work on the get endpoint - this should enable backwards compatibility

Please include:

- a summary of the changes

## Proposed version

- [] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


odrakes-cpr:
get request 
Screenshot 2024-12-23 at 13 38 07

odrakes-cpr:
Screenshot 2025-01-06 at 10 52 21


Refactored family/search/query params function

app/model/family.py
```diff
@@ -40,6 +41,7 @@ class FamilyWriteDTO(BaseModel):
     title: str
     summary: str
     geography: str
+    geographies: list[str]
```
katybaulch, line 44: If this PR is just for adding support for /get, do we need to update the FamilyWriteDTO here?

app/repository/family.py
```diff
@@ -199,25 +256,27 @@ def search(
             search.append(Family.description.ilike(term))
 
     if geography is not None:
-        geography_filter = or_(
-            *[(Geography.display_value == g.title()) for g in geography]
-        )
-        search.append(geography_filter)
+        geographies = [geo.capitalize() for geo in geography]
```
katybaulch, line 259: `capitalize` only capitalises the first letter of a string, not the whole string. Is that the behaviour we want?

app/repository/family.py
```diff
@@ -199,25 +256,27 @@ def search(
...
         query = _get_query(db).filter(condition)
         if org_id is not None:
             query = query.filter(Organisation.id == org_id)
+
         found = (
             query.order_by(desc(Family.last_modified))
             .limit(search_params["max_results"])
             .all()
         )
+
```
katybaulch, line 279: Nice formatting update!

tests/helpers/family.py
```diff
@@ -87,6 +89,7 @@ def create_family_write_dto(
     title: str = "title",
     summary: str = "summary",
     geography: str = "CHN",
+    geographies: list[str] = ["BRB", "BHS"],
```
katybaulch, line 92: Probably want to make the list ["CHN", "BRB", "BHS"] as the geography value is "CHN" and your logic says geography should be the first value in the geographies list

app/model/family.py
odrakes-cpr: ahh thanks for reminding me about this, sadly everything is interconnected with this PR - even though its a get there is a test for the [update endpoint ](https://github.com/climatepolicyradar/navigator-admin-backend/blob/c49a7316eb6adef4f0d9699beb7967a014df4ad9/tests/unit_tests/routers/family/test_update_family.py#L13) where we update the model and then I presume the FamilyReadDTO kicks in when the updated resource is sent back to the endpoint.  So i had to update these across the board so theres is no mismatch in whats going into the db

app/repository/family.py
odrakes-cpr: have removed for the current implimentation but yes that was the plan, probably need to update also for countries with more than one word but thats by the by now.

app/repository/family.py
```diff
@@ -67,6 +86,111 @@ def _family_to_dto(
         title=str(fam.title),
         summary=str(fam.description),
         geography=str(geo_value.value),
+        geographies=[str(geo_value.value)],
```
odrakes-cpr, line 89: is it worth using the geo_subquery function here to get the associated geographies, or leave that for another pr when we update the endpoint. probably the latter ?

app/model/family.py
katybaulch: Understood!

app/repository/family.py
katybaulch: The frontend will only need a list of the ISO codes associated with the family -- so whatever is the path of least resistance. Key is - you should be able to deploy this without anything breaking.

katybaulch (approved):
lgtm, nice work!!

PR #274, opened 2025-01-02
Feature/app 47 read corpus type information via admin service

katybaulch:
# Description

Add /all and /get routes for corpus types in the admin service.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/api/api_v1/routers/corpus_type.py
```diff
@@ -0,0 +1,58 @@
...
+    try:
+        return corpus_type_service.all(request.state.user)
+    except RepositoryError as e:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=e.message
+        )
+
+
+@corpus_types_router.get(
+    "/corpus-types/{corpus_type_name}",
```
jamesgorrie, line 34: This _feels_ like a relatively unsafe to access the data. Is this unique? Maybe an ID might be nicer?

tests/integration_tests/corpus/test_update.py
```diff
@@ -177,7 +177,7 @@ def test_update_corpus_non_super_admin(client: TestClient, admin_user_header_tok
     )
     assert response.status_code == status.HTTP_403_FORBIDDEN
     data = response.json()
-    assert data["detail"] == "User admin@cpr.org is not authorised to UPDATE a CORPORA"
+    assert data["detail"] == "User admin@cpr.org is not authorised to UPDATE a CORPUS"
```
jamesgorrie, line 180: Just double checking - we don't have anything that reads this and uses it for anything? Assuming it's just for logging.

tests/integration_tests/corpus_type/test_get.py
```diff
@@ -0,0 +1,70 @@
...
+    setup_db,
+)
+
+EXPECTED_CORPUS_TYPE_KEYS = ["name", "description", "metadata"]
+
+
+@pytest.mark.parametrize(
+    "expected_corpus_type",
+    [EXPECTED_CCLW_CORPUS, EXPECTED_UNFCCC_CORPUS],
+)
```
jamesgorrie, line 18: Nice. Wasn't aware you could do table based testing with a macro. Is there a benefit to this over just running a for loop [like here](https://github.com/climatepolicyradar/navigator-admin-backend/pull/266/files#diff-5adf0223ef05f4b9b3cc0b0b858726a7150441fb4db001972cc8d158dc8e3393R58-R74)?

[Looking at it](https://docs.pytest.org/en/stable/how-to/parametrize.html) - I think the way you've introduced has better test labelling.

tests/integration_tests/corpus_type/test_get.py
```diff
@@ -0,0 +1,70 @@
...
+    response = client.get(
+        f"/api/v1/corpus-types/{expected_corpus_type['corpus_type_name']}",
+        headers=superuser_header_token,
+    )
+    assert response.status_code == status.HTTP_200_OK
+    data = response.json()
+
+    assert set(EXPECTED_CORPUS_TYPE_KEYS).symmetric_difference(set(data.keys())) == set(
+        []
+    )
```
jamesgorrie, line 32: I think you can just use a simple `==` check on sets e.g.
https://onecompiler.com/python/434vgx9ja

jamesgorrie (commented):
Looks good, but I just wonder about using the `corpus_type_name` as the identifier.

I don't think this API is used outside `navigator-admin` ecosystem, so it shouldn't expose a breaking change outside of this scope, so maybe not a problem?

tests/integration_tests/corpus_type/test_get.py
katybaulch: I'm not sure performance wise, but it's more Pythonic, has better test labelling like you said, and helps make tests more reusable and isolated :)

app/api/api_v1/routers/corpus_type.py
katybaulch: Corpus types don't have IDs believe it or not 😬 The name is the primary key.

tests/integration_tests/corpus_type/test_get.py
katybaulch: Great spot - that's a hangover from some code I copied :)

katybaulch:
> Looks good, but I just wonder about using the `corpus_type_name` as the identifier.
> 
> I don't think this API is used outside `navigator-admin` ecosystem, so it shouldn't expose a breaking change outside of this scope, so maybe not a problem?


Corpus types don't have IDs believe it or not 😬 The name is the primary key. We won't be allowing the corpus type name to be changed in the admin service (at least at the moment) - just the corpus type description.

jamesgorrie:
> We won't be allowing the corpus type name to be changed in the admin service

Feels like something we should get to sooner rather than later - minting and IDing our objects. This is probably something that should percolate through the system too. 

But that's something that will need a more abstract design, outside the scope of this PR.

jamesgorrie (approved):
Fab

katybaulch:
> > We won't be allowing the corpus type name to be changed in the admin service
> 
> Feels like something we should get to sooner rather than later - minting and IDing our objects. This is probably something that should percolate through the system too.
> 
> But that's something that will need a more abstract design, outside the scope of this PR.

Yeah very much agreed!

PR #276, opened 2025-01-06
Feature/pdct 1722 fix corpus import id generator

katybaulch:
# Description

- Allow corpus import ID to be passed on create
- Add validation for corpus import ID
- Add test case

Note: The API remains dumb in -how- the corpus import IDs are generated for now. Instead, we will make the frontend build the desired import ID based on the options the user selects, and then we will use the backend to validate that the import ID is valid. 

This is because corpus import IDs aren't currently standardised entirely. For example, the below are all examples of corpus import IDs:
- UNFCCC.corpus.i00000001.n0000
- MCF.corpus.GCF.n0000
- MCF.corpus.GCF.Guidance

The matching frontend PR for this ticket will allow the user to build the corpus import ID using the following criteria:
- The import ID will contain 4 parts
- The first part will contain either the organisation name (not the org display name) or the organisation type
- The second part will always be 'corpus'
- The third part will be the organisation name if the first part contains the organisation type, otherwise the third part will contain a numbered index
- The fourth part will either be a numbered index as default, or free text that the user can input

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/model/corpus.py
+++ b/app/model/corpus.py
@@ -37,6 +37,7 @@ class CorpusWriteDTO(BaseModel):
 class CorpusCreateDTO(BaseModel):
     """Representation of a Corpus."""
 
+    import_id: Optional[str] = None
     title: str
     description: str
     corpus_text: Optional[str]
--- a/app/repository/config.py
+++ b/app/repository/config.py
@@ -47,7 +47,12 @@ def _to_corpus_data(row) -> CorpusData:
         description=row.description,
         corpus_type=row.corpus_type,
         corpus_type_description=row.corpus_type_description,
-        organisation={"name": row.org_name, "id": row.org_id},
+        organisation={
+            "name": row.org_name,
+            "id": row.org_id,
+            "display_name": row.org_display_name,
+            "type": row.org_type,
+        },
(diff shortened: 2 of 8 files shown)


odrakes-cpr:
this might be handle by the get distinct org options function or by sql itself (presuming import ids are unique entries) but are we handling instances where a user might create an import that already exists in the db

katybaulch:
> this might be handle by the get distinct org options function or by sql itself (presuming import ids are unique entries) but are we handling instances where a user might create an import that already exists in the db

Handled at the repository layer - the create won't happen if the primary key already exists in the table and a RepositoryError will be raised.

odrakes-cpr (approved):
sweet as

PR #278, opened 2025-01-08
Feature/pdct 1780 support multiple geos in update

odrakes-cpr:
# Description

Update the update endpoint to allow for the editing of multiple geographies.
- Added a separate check to the update intention which determines if the list of geo ids from the iso codes in the payload is diff from the associted family geo ids.
- Added a validation step to the service layer that checks if the iso code in the payload exists in the db

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

Updated the integration tests and the service / repository layer tests where appropriate

## Reviewer Checklist

- [ ] No FIXMEs remain


app/repository/family.py
```diff
@@ -113,6 +114,17 @@ def _update_intention(
         .geography_id
         != geo_id
     )
+
+    update_geographies = False
+    if geo_ids:
```
katybaulch, line 119: Can we explicitly make this `if geo_ids == []` please?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "admin_backend"
-version = "2.17.27"
+version = "2.17.29"
```
katybaulch, line 3: This will need bumping to `2.17.30`
```suggestion
version = "2.17.30"
```

tests/mocks/services/family_service.py
```diff
@@ -59,6 +59,7 @@ def mock_update_family(
                 data.title,
                 data.summary,
                 data.geography,
+                data.geographies,  # type: ignore
```
katybaulch, line 62: Cheeky - why are we ignoring the type here?

tests/unit_tests/service/geography/test_geography_service.py
```diff
@@ -16,3 +16,26 @@ def test_geo_service_validate_raises_when_invalid(
...
+        "One or more of the following geography values are invalid: CHN, USA, AGO"
+    )
+    assert e.value.message == expected_msg
+    assert geography_repo_mock.get_ids_from_values.call_count == 1
+
+
+def test_geo_service_gets_ids_from_repo(
+    geography_repo_mock,
+):
+    result = geography_service.get_ids(None, ["CHN", "USA"])  # type: ignore
```
katybaulch, line 39: Why are we ignoring the type here? Remember, casting is your friend :)

katybaulch (changes requested):
Nice work! Couple of comments 💬

tests/mocks/services/family_service.py
odrakes-cpr: ahh this was leftover, when i was testing the FamilyReadDTO, but those changes from that other PR have been merged in so i have removed.

app/model/family.py
```diff
@@ -41,7 +41,9 @@ class FamilyWriteDTO(BaseModel):
     title: str
     summary: str
     geography: str
-    geographies: list[str]
+    geographies: Optional[list[str]] = (
+        None  # remove default once implemented on the frontend
```
katybaulch, line 45: Can we add a TODO here please with the ticket number?

app/repository/family.py
```diff
@@ -238,6 +239,18 @@ def _update_intention(
         .geography_id
         != geo_id
     )
+
+    update_geographies = False
+    # TODO: remove this conditional once multi-geography support is implemented on the frontend
```
katybaulch, line 244: Ticket number on the TODO please :)

app/repository/family.py
```diff
@@ -475,6 +498,39 @@ def update(db: Session, import_id: str, family: FamilyWriteDTO, geo_id: int) ->
             )
             db.add(new_collection)
 
+    # Update geographies if geographies have changed.
+    if update_geographies:
+        original_geographies = set(
+            [
+                fg.geography_id
```
katybaulch, line 505: We should probably pull this into it's own function so we can reuse it and improve maintainability :)

app/repository/family.py
```diff
@@ -475,6 +498,39 @@ def update(db: Session, import_id: str, family: FamilyWriteDTO, geo_id: int) ->
...
+                for fg in db.query(FamilyGeography).filter(
+                    FamilyGeography.family_import_id == import_id
+                )
+            ]
+        )
+
+        cols_to_remove = set(original_geographies) - set(geo_ids)
+
+        for col in cols_to_remove:
+            result = db.execute(
```
katybaulch, line 515: Again - can we move this into its own function? We could have a function for removing old geographies that we no longer want `remove_old_geographies`, one for adding new geographies `add_new_geographies`, and we can wrap both of those functions in a third function called `update_geographies`. This should make this function more maintainable but also help make it more testable

app/repository/geography.py
```diff
@@ -5,4 +5,25 @@
...
+    """
+    Fetch IDs for multiple geographies based on their iso values.
+
+    :param Session db: Database session.
+    :param list[str] geo_strings: A list of geography iso values to look up.
+    :return list[int]: A list of IDs corresponding to the provided geography values.
+    """
+    return [
+        geography.id
+        for geography in db.query(Geography).filter(Geography.value.in_(geo_strings))
```
katybaulch, line 28: do we need to call .all() on this statement?

app/repository/family.py
odrakes-cpr: @katybaulch there is a helper file, would it be more suitable there ?

katybaulch: umm i'd say no as the helpers i don't think are specific to an entity (e.g., family/geography)

app/repository/family.py
```diff
@@ -475,6 +499,10 @@ def update(db: Session, import_id: str, family: FamilyWriteDTO, geo_id: int) ->
             )
```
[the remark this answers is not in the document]
katybaulch: Not in the wrapper function _ i'm an idiot

app/repository/family.py
```diff
@@ -655,3 +683,79 @@ def count(db: Session, org_id: Optional[int]) -> Optional[int]:
...
+    """
+    cols_to_add = set(geo_ids) - set(original_geographies)
+
+    for col in cols_to_add:
+        db.flush()
+        new_geography = FamilyGeography(
+            family_import_id=import_id,
+            geography_id=col,
+        )
+        db.add(new_geography)
```
katybaulch, line 737: Add raising RepoError in this function

katybaulch (changes requested):
Missing a RepoError raise for geography add but otherwise super great, well done!!!! 🦖

katybaulch (approved):
very nice!!!!

PR #285, opened 2025-01-27
Fix family all endpoint running slow due to an intensive query

odrakes-cpr:
# Description

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


odrakes-cpr:
https://linear.app/climate-policy-radar/issue/APP-183/fix-slow-family-endpoint-in-the-admin-service

app/repository/family.py
```diff
@@ -105,6 +105,180 @@ def _get_query(db: Session) -> Query:
...
+def _get_query_all_search_endpoint(
+    db: Session, raw_sql_query: str, query_params: Optional[dict] = {}
+) -> Query:
+    result = db.execute(text(raw_sql_query), query_params)
+    return result.mappings().fetchall()
+
+
+def construct_raw_sql_query(org_id=None, filters=None, filter_params=None):
+    main_sql_query = """
+        SELECT
```
katybaulch, line 117: suggestion: can we move this into a separate file for maintainability/readability purposes please?

app/repository/family.py
```diff
@@ -192,6 +366,9 @@ def _update_intention(
     )
 
 
+# NOTE ONLY UPDATE THE ALL ENDPOINT DUPLICATE THE SEARCH AND ALL ENDPOINT
```
annaCPR, line 369: issue (non-blocking): Not sure I understand this comment. Is it possible to reword it to be clearer?

annaCPR (commented):
suggestion: Would you mind documenting in the PR description why and how this solution fixes the problem we had with the endpoint? Just so that I don't have to try to figure it out from the SQL :-P

odrakes-cpr:
Resulting query - 
Screenshot 2025-01-29 at 12 27 44

odrakes-cpr:
> suggestion: Would you mind documenting in the PR description why and how this solution fixes the problem we had with the endpoint? Just so that I don't have to try to figure it out from the SQL :-P

Sure hopefully that has helped.

app/repository/family.py
```diff
@@ -249,46 +308,69 @@ def search(
...
+        if "summary" in search_params:
             term = f"%{escape_like(search_params['summary'])}%"
-            search.append(Family.description.ilike(term))
+            conditions.append("f.description ILIKE :summary")
+            params["summary"] = term
 
     if geography is not None:
-        import_ids = _get_family_import_ids_by_geographies(db, geography)
-        search.append(Family.import_id.in_(import_ids))
+        family_geographies = _get_family_geographies_by_display_values(db, geography)
```
jamesgorrie, line 332: suggestion: similar to my other comment - should this just be inline?

If a method doesn't add an abstraction that's reusable or makes it testable - I often question if it should be an abstraction at all.

If anything it just adds a step for our future selves to jump through and having to ask the question of what's the abstraction for if we wanted to refactor.

app/repository/helpers.py
```diff
@@ -173,3 +173,146 @@ def perform_family_geographies_update(db: Session, import_id: str, geo_ids: list
...
+    main_sql_query += """
+    GROUP BY
+        f.import_id, f.title, geography_subquery.geography_values,
+        family_documents_subquery.document_ids, family_events_subquery.event_ids,
+        fm.family_import_id, c.import_id, o.id, slug_subquery.slugs
+    ORDER BY f.last_modified DESC
+    LIMIT :max_results
+    """
+
+    return main_sql_query, query_params
```
jamesgorrie, line 318: question: This is quite complicated, but I am assuming the integration tests are catching that it's giving us what we need?

jamesgorrie (approved):
Really nice investigative work - I wish we had some metrics to show the improvement in real time.

Might be worth getting current local metrics and compare them so we can see the improvement.

app/repository/helpers.py
odrakes-cpr: Yes definitely - will have a think if there is anything else I could add.

jamesgorrie: Not sure we need to - just double checking!

PR #286, opened 2025-01-28
Add /get and /all for organisations

katybaulch:
# Description

Add /get and /all routes for organisations.

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/organisation.py
+++ b/app/api/api_v1/routers/organisation.py
@@ -1,0 +1,59 @@
+import logging
+from typing import Union
+
+from fastapi import APIRouter, HTTPException, status
+
+from app.errors import RepositoryError, ValidationError
+from app.model.organisation import OrganisationReadDTO
+from app.service import organisation as organisation_service
+
+organisations_router = APIRouter()
+
+_LOGGER = logging.getLogger(__name__)
+
+
+@organisations_router.get(
+    "/organisations",
+    response_model=list[OrganisationReadDTO],
+)
+async def get_all_organisations() -> list[OrganisationReadDTO]:
+    """Retrieve all organisations.
+
+    :raises HTTPException: If the corpus type is not found.
+    :return CorpusTypeReadDTO: The requested corpus type.
+    """
+    try:
+        return organisation_service.all()
+    except RepositoryError as e:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=e.message
+        )
+
+
+@organisations_router.get(
+    "/organisations/{organisation}",
+    response_model=OrganisationReadDTO,
+)
+async def get_organisation(organisation: Union[int, str]) -> OrganisationReadDTO:
+    """Retrieve a specific organisation by its name.
+
+    :param Union[int, str] organisation: The ID or name of the
+        organisation to retrieve.
+    :raises HTTPException: If the organisation is not found.
+    :return OrganisationReadDTO: The requested organisation.
+    """
+    try:
+        org = organisation_service.get(organisation)
+    except ValidationError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=e.message)
+    except RepositoryError as e:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=e.message
+        )
+
+    if org is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Organisation not found: {organisation}",
+        )
+    return org
--- a/app/api/api_v1/routers/organisation.py
+++ b/app/api/api_v1/routers/organisation.py
@@ -1,5 +1,4 @@
 import logging
-from typing import Union
 
 from fastapi import APIRouter, HTTPException, status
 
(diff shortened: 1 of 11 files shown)


app/api/api_v1/routers/organisation.py
```diff
@@ -0,0 +1,59 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=e.message
+        )
+
+
+@organisations_router.get(
+    "/organisations/{organisation}",
+    response_model=OrganisationReadDTO,
+)
+async def get_organisation(organisation: Union[int, str]) -> OrganisationReadDTO:
```
jamesgorrie, line 37: question: why would we need to use the name over the ID? Having the 2 means we need to support this going forward - and overloads this URL.

This will also lead us to have to manage if 2 orgs have the same name.

suggestion: Following the rest principle - each resource should be uniquely identifiable - could we have something like I would expect:

`/organisations/{id}` -> fetches one.

And if we needed to fetch by name something like:
`/organisations?q={organisation_name}` -> finds an org with that name.

jamesgorrie (commented):
mainly a question around why we need the endpoint to use the org name and ID as the fetch param - and a suggestion that maybe we support the fetching by name in a different way.

app/api/api_v1/routers/organisation.py
katybaulch: Extremely fair - have made this get by ID only.

app/api/api_v1/routers/organisation.py
```diff
@@ -0,0 +1,57 @@
...
+    try:
+        return organisation_service.all()
+    except RepositoryError as e:
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=e.message
+        )
+
+
+@organisations_router.get(
+    "/organisations/{organisation}",
```
jamesgorrie, line 33:
```suggestion
    "/organisations/{organisation_id}",
```

nit: following docs - maybe `_id` is more explicit https://fastapi.tiangolo.com/tutorial/path-params/

app/repository/organisation.py
```diff
@@ -20,3 +28,71 @@ def get_distinct_org_options(db: Session) -> list[str]:
...
+        org = db.query(Organisation).filter(Organisation.id == org_id).one_or_none()
+
+    except MultipleResultsFound as e:
+        _LOGGER.error(e)
+        raise RepositoryError(e)
+
+    return _org_to_dto(org) if org is not None else None
+
+
+def get_by_name(db: Session, org_name: str) -> Optional[OrganisationReadDTO]:
```
jamesgorrie, line 83: question: Is this used somewhere? I can't see where if it is?

app/service/organisation.py
```diff
@@ -1,11 +1,45 @@
...
 
 
 def get_id_from_name(db: Session, org_name: str) -> int:
     id = organisation_repo.get_id_from_name(db, org_name)
     if id is None:
         raise ValidationError(f"The organisation name {org_name} is invalid!")
     return id
+
+
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
```
jamesgorrie, line 22: question: I see we use this in a few places - just wondering why?

tests/integration_tests/setup_db.py
```diff
@@ -304,6 +304,22 @@ class DBEntry(TypedDict):
     "metadata",
 ]
 
+EXPECTED_NUM_ORGS = 3
```
jamesgorrie, line 307: question: A little lost - why is this three?

jamesgorrie (approved):
non-blocking questions is all - looks great!

app/service/organisation.py
katybaulch: enforce strict typing like your favourite typed languages -- if a param is passed that is of the wrong type or the return type is of the wrong type, it raises an error!

jamesgorrie: It sounds like `arbitrary_types_allowed` would loosen that though?

katybaulch: I think that is required as we haven't implemented `__get_pydantic_core_schema__` on our models - and without it a decorator without arbitrary types or with them set to False raises an error.

tests/integration_tests/setup_db.py
katybaulch: Because in our test data we set up the database with before our integration tests contains 3 organisations; CCLW and UNFCCC and another test org.

app/repository/organisation.py
katybaulch: Removed.

app/api/api_v1/routers/organisation.py
katybaulch: Done

app/service/organisation.py
jamesgorrie: Thanks! Sorry for getting deep into irrelevant pydantic shizzle.

tests/integration_tests/setup_db.py
jamesgorrie: I thought as much but was battling to see where... thanks!

PR #287, opened 2025-01-30
Feature/app 183 add logging to search endpoint

odrakes-cpr:
# Description

Add logging to the search endpoint, we want to not only measure the time it takes to run a query but use this metric to determine if a fix has made a viable improvement for our users. 

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/repository/family.py
```diff
@@ -288,7 +289,12 @@ def search(
             raise TimeoutError
         raise RepositoryError(e)
 
-    return [_family_to_dto(db, f) for f in found]
+    start_time = time.time()
+    results = [_family_to_dto(db, f) for f in found]
+    end_time = time.time()
+    print(f"Search Query Execution Time: {end_time - start_time} seconds")
```
jamesgorrie, line 295: Should we be logging rather than printing [like we do elsewhere](https://github.com/search?q=repo%3Aclimatepolicyradar%2Fnavigator-admin-backend%20logger&type=code)?

I'm not sure the print will make it out to the logs.

jamesgorrie (commented):
Note on logging - also wondered as a secondary step we might just [add a middleware that does this](https://fastapi-utils.davidmontague.xyz/user-guide/timing-middleware/)?

app/repository/family.py
katybaulch: The custom logger might swallow the `print` statement, so I think `logging` might be safer!

odrakes-cpr:
> Note on logging - also wondered as a secondary step we might just [add a middleware that does this](https://fastapi-utils.davidmontague.xyz/user-guide/timing-middleware/)?

Updated accordingly.

It adds `response_time_ms`

```
{"written_at": "2025-02-03T10:35:16.587Z", "written_ts": 1738578916587310000, "type": "request", "correlation_id": "8ee06d94-e21a-11ef-b24c-0242ac120003", "remote_user": "-", "request": "/api/v1/families/", "referer": "-", "x_forwarded_for": "-", "protocol": "HTTP/1.1", "method": "GET", "remote_ip": "172.18.0.1", "request_size_b": -1, "remote_host": "172.18.0.1", "remote_port": 58306, "request_received_at": "2025-02-03T10:35:14.574Z", "response_time_ms": 2012, "response_status": 200, "response_size_b": "835470", "response_content_type": "application/json", "response_sent_at": "2025-02-03T10:35:16.587Z"}
```

app/main.py
```diff
@@ -138,6 +136,7 @@ async def root():
     return {"message": "CPR Navigator Admin API v1"}
 
 
+# trunk-ignore(trunk/DS002)
```
katybaulch, line 139: question: I can't see what this ignore is ignoring. 

When it is removed, there is no trunk reported error locally?

katybaulch (approved):
great find - love the middleware!

app/main.py
odrakes-cpr: `Possible binding to all interfaces.`

I believe its complaining about the host port a few lines down, trunk is complaining that it could be a security-related issue if we expose our FastAPI to all network interfaces. I think that we bind docker containers to `0.0.0.0` as standard practice but I actually left in accidentally when I was trying to figure out why trunk was complaining about this file. I shall remove!

odrakes-cpr: ok, I will remove separately, because sadly it screws up my formatting again - if i try update this file.

PR #301, opened 2025-03-03
Feature/app 287 improve our logging

annaCPR:
# Description

- remove duplicated and/or unhelpful logs
- add more details to log messages or rephrase where needed to make it clearer as to what happened
- add debug logs to bulk update that show what has changed and the values being updated
- explicitly set log level to INOF in modules where we log either INFO or DEBUG level logs so that they show up and we can also use an env var to easily switch to DEBUG if needed

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain

### DIFF
--- a/app/api/api_v1/routers/bulk_import.py
+++ b/app/api/api_v1/routers/bulk_import.py
@@ -33,8 +33,6 @@ async def get_bulk_import_template(corpus_type: str) -> Json:
     :return Json: json representation of bulk import template.
     """
 
-    _LOGGER.info(f"Creating template for corpus type: {corpus_type}")
-
     try:
         return {
             "collections": [get_collection_template()],
@@ -81,13 +79,13 @@ async def bulk_import(
             "message": "Bulk import request accepted. Check Cloudwatch logs for result."
         }
     except ValidationError as e:
-        _LOGGER.error(e.message, exc_info=True)
+        _LOGGER.error(e.message)
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=e.message)
     except HTTPException as e:
-        _LOGGER.error(e, exc_info=True)
+        _LOGGER.error(e)
         raise e
     except Exception as e:
-        _LOGGER.error(e, exc_info=True)
+        _LOGGER.error(e)
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=str(e)
         )
--- a/app/service/token.py
+++ b/app/service/token.py
@@ -72,9 +72,10 @@ def decode(token: str) -> UserContext:
     """
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-    except jwt.PyJWTError:
-        _LOGGER.exception("Error when decoding")
-        raise TokenError("Payload cannot be decoded")
+    except jwt.PyJWTError as e:
+        msg = f"Error when decoding token: {e}"
+        _LOGGER.exception(msg)
+        raise TokenError(msg)
 
     email: Optional[str] = payload.get("email")
     if email is None:
--- a/app/api/api_v1/query_params.py
+++ b/app/api/api_v1/query_params.py
@@ -1,9 +1,11 @@
 import logging
+import os
 from typing import Union, cast
 
 from fastapi import HTTPException, status
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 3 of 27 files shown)


app/api/api_v1/routers/bulk_import.py
```diff
@@ -81,13 +79,13 @@ async def bulk_import(
             "message": "Bulk import request accepted. Check Cloudwatch logs for result."
         }
     except ValidationError as e:
-        _LOGGER.error(e.message, exc_info=True)
+        _LOGGER.error(e.message)
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=e.message)
     except HTTPException as e:
-        _LOGGER.error(e, exc_info=True)
+        _LOGGER.error(e)
```
katybaulch, line 85: Question (non blocking): Are we happy suppressing the traceback here?
If we want the traceback, we can use 'logger.exception(e.message)' instead.

katybaulch (changes requested):
looking good! please can we decide whether we want to use logger.error or logger.exception though, as we seem to be using both inconsistently. I'd argue that if we are trying to improve our logging, we should use logger.exception over logger.error so we get the full traceback to help us debug (if needed) - but happy to be convinced otherwise! ⭐

app/api/api_v1/routers/bulk_import.py
annaCPR: See above

app/api/api_v1/routers/bulk_import.py
```diff
@@ -81,13 +79,13 @@ async def bulk_import(
...
         raise e
     except Exception as e:
-        _LOGGER.error(e, exc_info=True)
+        _LOGGER.error(e)
```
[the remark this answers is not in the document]
annaCPR: See above

annaCPR:
> looking good! please can we decide whether we want to use logger.error or logger.exception though, as we seem to be using both inconsistently. I'd argue that if we are trying to improve our logging, we should use logger.exception over logger.error so we get the full traceback to help us debug (if needed) - but happy to be convinced otherwise! ⭐

@katybaulch do you mean that for the entire repo or just these specific cases that you pointed out?

app/service/token.py
```diff
@@ -72,9 +72,10 @@ def decode(token: str) -> UserContext:
     """
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-    except jwt.PyJWTError:
-        _LOGGER.exception("Error when decoding")
-        raise TokenError("Payload cannot be decoded")
+    except jwt.PyJWTError as e:
+        msg = f"Error when decoding token: {e}"
+        _LOGGER.exception(msg)
```
katybaulch, line 77: suggestion: use either exception here or error -- but we should be consistent with the rest of the repo

katybaulch:
> > looking good! please can we decide whether we want to use logger.error or logger.exception though, as we seem to be using both inconsistently. I'd argue that if we are trying to improve our logging, we should use logger.exception over logger.error so we get the full traceback to help us debug (if needed) - but happy to be convinced otherwise! ⭐
> 
> @katybaulch do you mean that for the entire repo or just these specific cases that you pointed out?

In the files you've changed :) I've commented on another file where we are using exception instead of error rather than vice versa

app/service/token.py
annaCPR: @katybaulch we don't have a consistent approach in the rest of the repo as far as I can see (we use both) so which one would you like me to use here? (_LOGGER.exception was the original implementation - I only refactored the message)

app/api/api_v1/routers/bulk_import.py
```diff
@@ -81,13 +79,13 @@ async def bulk_import(
             "message": "Bulk import request accepted. Check Cloudwatch logs for result."
         }
     except ValidationError as e:
-        _LOGGER.error(e.message, exc_info=True)
+        _LOGGER.error(e.message)
```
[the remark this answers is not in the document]
annaCPR: Will change it to exception log then.

odrakes-cpr (approved):
look slike youve addressed everything

PR #307, opened 2025-03-11
Bump db-client with new geographies

jamesgorrie:
# Description

- [updates db-client](https://github.com/climatepolicyradar/navigator-db-client/compare/v3.8.29...v3.9.1), with the main update being adding more geographies
- filters the geographies returned by the config endpoint to exclude these new geographies

This wouldn't affect the admin interface either way as we don't read deeper than 1 level, but it was returning a lot of extra data down the pipes that was unused.

This maintains the functionality of returning those subdivisions later if we want.

## Proposed version

- [x] Patch


tests/integration_tests/setup_db.py
```diff
@@ -517,6 +517,7 @@ def _setup_collection_data(
                 import_id=data["import_id"],
                 title=data["title"],
                 description=data["description"],
+                valid_metadata={},
```
jamesgorrie, line 520: This was needed due to a constraint added in the db-client. CC @annaCPR

odrakes-cpr (approved):
v nice

annaCPR (approved):
lgtm

PR #314, opened 2025-03-20
Fix/app 404 bug cannot create collections via the admin service

annaCPR:
# Description
- make collection metadata field optional on the CollectionCreate and CollectionWrite models since not all corpora will have metadata for collections

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] **DB_CLIENT DEPENDENCY IS ON THE LATEST VERSION**
- [ ] No FIXMEs remain

### DIFF
--- a/app/model/collection.py
+++ b/app/model/collection.py
@@ -25,7 +25,7 @@ class CollectionWriteDTO(BaseModel):
     title: str
     description: str
     organisation: str
-    metadata: Json
+    metadata: Optional[Json] = {}
 
 
 class CollectionCreateDTO(BaseModel):
--- a/app/model/collection.py
+++ b/app/model/collection.py
@@ -34,4 +34,4 @@ class CollectionCreateDTO(BaseModel):
     import_id: Optional[str] = None
     title: str
     description: str
-    metadata: Json
+    metadata: Optional[Json] = {}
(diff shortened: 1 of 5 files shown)


app/model/collection.py
```diff
@@ -34,4 +34,4 @@ class CollectionCreateDTO(BaseModel):
     import_id: Optional[str] = None
     title: str
     description: str
-    metadata: Json
+    metadata: Optional[Json] = {}
```
odrakes-cpr, line 37: question: do we have the metadata set as nullable in the db-client or db also, trying to figure out how it works

odrakes-cpr (approved):

app/model/collection.py
annaCPR: It is required in the db and SQLAlchemy model and the idea is to default to {} for corpora that don't have metadata on collections. But we only default the value here (i.e. in the BE models), is that right? It is possible for the metadata value to be explicitly set to None but that would result in a db error

PR #324, opened 2025-04-24
fix: filter geo deletion by family_import_id

jamesgorrie:
# Description
- filters the geo deletion on `family.update` by `family.import_id`
- removes unused util code as it's confusing

This fixes the issue of the :ghost: `family_geography` table deletions.

part of: https://linear.app/climate-policy-radar/issue/APP-538/data-loss-to-family-geography-table

## Proposed version

- [x] Patch


katybaulch (approved):
lgtm

PR #327, opened 2025-04-30
Improve bulk import logs

annaCPR:
# Description
- change the log level from DEBUG to INFO when we detect changes to update as we found that it's always good to see what is being updated
- round up the timings of how long each step took 
- add a log message if there was no data to import (which can happen with automatic updates) just to make it clearer what is happening (and also skip saving the input and the result to s3 in that case as it would be empty anyway)

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/service/bulk_import.py
```diff
@@ -480,6 +476,10 @@ def import_data(
         db.commit()
 
         if any([collection_data, family_data, document_data, event_data]):
+            import_uuid = uuid4()
+            upload_bulk_import_json_to_s3(
+                f"{import_uuid}-request", corpus_import_id, data
+            )
             upload_bulk_import_json_to_s3(
```
odrakes-cpr, line 483: quibble: assume github is playing tricks on me here, but you do have this upload function twice right

odrakes-cpr (approved):
lgtm: comment resolved

PR #328, opened 2025-04-30
Feature/app 282 data dumps of rds stores

odrakes-cpr:
# Description

Please include:

Create a dump of our navigator database.
Create a function that uploads that database to our bucket in s3
Create a function that deletes that dump file

Trigger these in the bulk import

## Proposed version

- [ ] Skip auto-tagging
- [ ] Patch
- [x] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [ ] No FIXMEs remain


app/clients/aws/s3bucket.py
```diff
@@ -130,4 +131,36 @@ def upload_bulk_import_json_to_s3(
...
 
+def upload_sql_db_dump_to_s3(dump_file: str) -> None:
+    """
+    Upload the database dump to S3 and clean up local file.
+
+    Args:
+        dump_file (str): Path to the dump file
+    """
+    s3_client = boto3.client("s3")
+    bucket_name = os.environ["DATABASE_DUMP_BUCKET"]
```
katybaulch, line 142: question: do we need to add this to our `.env.example` file?

app/clients/aws/s3bucket.py
```diff
@@ -130,4 +131,36 @@ def upload_bulk_import_json_to_s3(
...
+
+        _LOGGER.info(f"🚀 Uploading {dump_file} to S3 bucket {bucket_name}")
+
+        with dump_path.open("rb") as f:
+            s3_client.upload_fileobj(f, bucket_name, s3_key)
+
+        _LOGGER.info("🎉 Database Dump upload completed successfully")
+
+    except Exception as e:
+        _LOGGER.error(f"💥 Upload failed: {e}")
```
katybaulch, line 162: suggestion: you could use `_LOGGER.exception(msg)` here instead of including the traceback inline in the error message

app/clients/aws/s3bucket.py
```diff
@@ -130,4 +131,36 @@ def upload_bulk_import_json_to_s3(
...
+        raise
```
katybaulch, line 163: issue (non-blocking): bare `raise`. What are you raising here? The same exception? A different exception?

katybaulch: suggestion: I'd remove the try-except wrapper in this function and handle the error catching in the calling function - and log out the error there rather than doing this.

app/service/bulk_import.py
```diff
@@ -495,4 +507,8 @@ def import_data(
         db.rollback()
         end_message = f"💥 Bulk import for corpus: {corpus_import_id} has failed."
     finally:
+        try:
+            trigger_db_dump_upload_to_sql()
+        except Exception as e:
+            _LOGGER.error(f"⚠️ Failed to trigger DB dump upload: {e}", exc_info=True)
```
katybaulch, line 513: suggestion: you could use _LOGGER.exception(msg) here instead of including the traceback inline in the error message as well as also passing the full stack trace. I think this technically will print the same traceback twice?

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+    :param value: The Postgres parameter to validate
+    returns str: The validated parameter
+    :raises ValueError: If the parameter is invalid
+    """
+    if value is None or not re.match(r"^[\w.-]+$", value):
+        raise ValueError(f"Invalid Postgres parameter: {value}")
+    return value
+
+
+def delete_local_file(file_path: str) -> None:
```
katybaulch, line 31: suggestion: you could make this return a `bool`, and handle whether it successfully deleted the file using a conditional in the calling function rather than this try-except block with a bare `raise` (not good practice)

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+    :raises RuntimeError: If security checks fail or the operation times out.
+    :return str: The path to the generated SQL dump file.
+    """
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    dump_file = Path(f"navigator_dump_{timestamp}.sql")
+
+    if dump_file.exists():
+        raise RuntimeError(f"Dump file already exists: {dump_file}")
+
+    # pg_dump command
```
katybaulch, line 61: quibble: how can we make this comment more meaningful?

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+        _LOGGER.info(f"🚀 Starting database dump to {dump_file}")
+
+        result = subprocess.run(
+            cmd,
+            check=True,
+            env=env,
+            capture_output=True,
+            text=True,
+            shell=False,
+            timeout=300,
```
katybaulch, line 90: question: what is the unit of this?

katybaulch, line 90: issue: magic number! what is the unit of this? assuming millisecs:
```suggestion
            timeout_ms = 300
            timeout=timeout_ms,
```

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+        _LOGGER.info("✅ Database dump completed successfully")
+        if result.stdout:
+            _LOGGER.debug(f"pg_dump output: {result.stdout}")
+        return str(dump_file)
+
+    except subprocess.TimeoutExpired:
+        _LOGGER.error("⌛ Database dump timed out")
+        if dump_file.exists():
+            dump_file.unlink()
+        raise RuntimeError("Database dump timed out after 5 minutes")
```
katybaulch, line 104: issue: this doesn't seem to correspond to the timeout value set above - can we make these values match?

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+            _LOGGER.error(f"stderr: {e.stderr}")
+        if dump_file.exists():
+            dump_file.unlink()
+        raise
+
+    except Exception as e:
+        _LOGGER.error(f"⚠️ Unexpected error during database dump: {e}")
+        if dump_file.exists():
+            dump_file.unlink()
+        raise
```
katybaulch, line 118: question: are we raising the same exception here? if yes, we can use the `raise from e` syntax

app/clients/aws/s3bucket.py
annaCPR: I quite like it that the function encapsulates all of the behaviour, including logging error, it makes the code cleaner, especially as the bulk import service already does a lot. I would actually remove the try catch in the level above as we're catching and logging the same error twice and this is at the end of the bulk import process so just logging that something went wrong is enough we don't need to raise the exception again (if that makes sense)

app/service/bulk_import.py
annaCPR: This is the try/catch I was talking about above. We've already logged this error, haven't we?

tests/unit_tests/service/test_sql_data_dump.py
```diff
@@ -0,0 +1,228 @@
...
+    """Test timeout during database dump"""
+
+    mock_file = MagicMock()
+    mock_file.exists.side_effect = [False, True]  # First call: False, Second: True
+    mock_file.__str__ = MagicMock(return_value="test_dump_timeout_error.sql")
+    mock_path.return_value = mock_file
+
+    mock_run.side_effect = subprocess.TimeoutExpired("pg_dump", timeout=300)
+
+    with pytest.raises(RuntimeError, match="Database dump timed out after 5 minutes"):
```
katybaulch, line 137: question: should we also be testing the timeout value properly here? The '5 minutes' text is hardcoded, whereas about the timeout is set to 300 (I assume) seconds? What happens if someone changes the timeout value in the function we're testing? This test would pass when it should fail.

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+        if path.exists():
+            path.unlink()
+            _LOGGER.debug(f"Deleted local file: {file_path}")
+    except Exception as e:
+        _LOGGER.error(f"⚠️ Failed to delete file {file_path}: {e}")
+        raise
+
+
+def get_database_dump() -> str:
```
katybaulch, line 44: suggestion: make `timeout` an optional parameter here with a default of 300.
```suggestion
def get_database_dump(timeout_secs: int = 300) -> str:
```

app/service/database_dump.py
annaCPR: I'm not sure I agree about returning a bool... That still requires the caller to check the return value to decide if the action worked or not, rather than assuming it has if no exception was thrown. Seems like more effort for less clarity. If a function fails to do what it was written to do, it should throw. Also, exceptions give more context of what went wrong by default.

annaCPR (approved):
lgtm

katybaulch (dismissed):
nice one! a few issues, questions & suggestions - shout me if you want to talk about them :)

the reason I'm rejecting is that you have an issue with a function that could cause one of your test cases to falsely pass when it should fail. the timeout value isn't being tested properly - the '5 minutes' text is hardcoded, whereas about the timeout is set to 300 (I assume) seconds? What happens if someone changes the timeout value in the function we're testing? This test would pass when it should fail.

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+            cmd,
+            check=True,
+            env=env,
+            capture_output=True,
+            text=True,
+            shell=False,
+            timeout=300,
+        )
+
+        dump_file.chmod(0o600)
```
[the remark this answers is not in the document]
odrakes-cpr: Was just adding some extra protections here; 
Once the file has been created, set restrictive permissions so that it can't be accidentally modified, prevents other processes or users from reading the dump, if someone or another process gains access, they can't read the file

app/clients/aws/s3bucket.py
odrakes-cpr: I don't think so as we handle the configs in the `docker-compose-test.yml` and `docker-compose-dev.yml`, but will add it in for completeness incase someone doesn't use those.

app/clients/aws/s3bucket.py
katybaulch: in that case - we want to make the `raise` here not bare :)

app/service/database_dump.py
katybaulch: both approaches are subjective! my comment is a suggestion, I'm okay with it not being actioned:))

app/service/bulk_import.py
katybaulch: in that case, it'll log three(!) + times

app/service/database_dump.py
odrakes-cpr: will update the bare raise, but yes in agreement as I think I am leaning into the fail fast principle, handling errors upfront (and raising them lol) and ensuring it fails loudly rather than leaving it to the caller to handle the result.

app/service/bulk_import.py
odrakes-cpr: agreed, will remove.

app/clients/aws/s3bucket.py
odrakes-cpr: Not sure if I am crossing paths here, but I want to keep the error logging contained for the actual upload functionality (raising the exceptions appropriately) . the level above this is 

```
def trigger_db_dump_upload_to_sql() -> None:
    dump_file = get_database_dump()
    try:
        upload_sql_db_dump_to_s3(dump_file)
    finally:
        delete_local_file(dump_file)

```

which I want to keep because here the finally block ensures that delete_local_file(dump_file) is executed no matter what happens in upload_sql_db_dump_to_s3(dump_file). So even if an exception is raised during the upload process, the local file is deleted after the upload attempt (successful or not).

app/service/database_dump.py
```diff
@@ -0,0 +1,118 @@
...
+    """Delete a local file safely, logging any issues."""
+    path = Path(file_path)
+
+    try:
+        if path.exists():
+            path.unlink()
+            _LOGGER.debug(f"Deleted local file: {file_path}")
+    except Exception as e:
+        _LOGGER.error(f"⚠️ Failed to delete file {file_path}: {e}")
+        raise
```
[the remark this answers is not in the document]
odrakes-cpr: the function that calls the db dump would also call the upload so it would be responsible to handle both file related_errors and upload errors, I think I am pushing to keep the error context in the function that actually raises the error just to make it clear where exactly the failure occurred. Especially as there already is a lot going on in that bulk import file.

app/clients/aws/s3bucket.py
katybaulch: understood!

app/service/database_dump.py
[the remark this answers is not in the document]
katybaulch: ta, having the context for this in a comment like you have is super useful!

annaCPR (approved):
lgtm

PR #334, opened 2025-06-02
fix: close db connections

jamesgorrie:
# Description

- closes DB connections via `contextmanager`

[We seem to have a partially implemented way to do this](https://github.com/climatepolicyradar/navigator-admin-backend/blob/main/app/clients/db/session.py#L27-L55). 

I have chosen not to use this - as the refactor would be quite large. If we were going to do a refactor of this size - I would chose to use [FastAPIs Depends](https://fastapi.tiangolo.com/tutorial/dependencies/#create-a-dependency-or-dependable). This again is a larger refactor.

**_I would like to validate that this solves our connection pool size issue before embarking on any refactor_** 

## Proposed version

- [x] Patch


jamesgorrie:
towards: https://linear.app/climate-policy-radar/issue/APP-630/investigate-hypothesis-that-admin-service-backend-is-leaving-stray

tests/unit_tests/clients/db/test_session.py
```diff
@@ -0,0 +1,40 @@
...
+        # the session should still be open
+        assert getattr(db, "closed") is False
+
+    # Now we're outside the with-block, the session should be closed
+    assert mock_session.closed is True
+
+
+def test_get_db_closes_after_exception_is_thrown(monkeypatch):
+
+    import app.clients.db.session as db_session
```
annaCPR, line 30: question: Any reason we can't move this import to the top to avoid duplication?

jamesgorrie: @katybaulch The reason I have done this is to localise the `monkeypatch` rather than have it in an mutated state from tests before it.

jamesgorrie:
@annaCPR I've changed the implementation so that I can do this across the codebase incrementally.

So first:
- this PR: introduce the new session mananged `get_db_session`
- this PR: use this in the `with_database` method
- this PR: if somewhere uses with `with db_session.get_db() as db` => `with db_session.get_db_session() as db` <= this means that we should have these connections properly closed
- next PRs (incremental): convert `db = db_session.get_db` => `with db_session.get_db_session() as db`
- potential PRs: migrate to FastAPI `Depends`

So the main reason to do this is to start making sure we aren't leaking connection so we can get on with not having to keep worrying about RDS so much

jamesgorrie:
Closed as the route to a good refactor needs work

PR #335, opened 2025-06-03
Feature/app 343 tidy up duplicated event type taxonomy

annaCPR:
# Description
- add a metadata field to the Create and Write EventDtos
- remove event_metadata as a separate parameter to the create and update methods in the event repository layer
- remove any instances where the top level event_type subtaxonomy was used (as it has now been superceded by _event)
- remove functions that created event metadata based on the given event data and relevant taxonomy as we now expect the metadata to be passed in as part of the event dtos
- update db_client to the latest version which includes a new migration script to update 'Laws and Policies' and 'Intl. agreements' taxonomies to no longer have the top level "event_type" subtaxonomy

Please include:

- a summary of the changes

## Proposed version

- [ ] Skip auto-tagging
- [x] Patch
- [ ] Minor version
- [ ] Major version

Notes:

  brackets and the `x`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] GitHub workflow update
- [ ] Documentation update
- [ ] Remove legacy code
- [ ] Dependency update

## How Has This Been Tested?

## Reviewer Checklist

- [x] **DB_CLIENT DEPENDENCY IS ON THE LATEST VERSION**
- [ ] No FIXMEs remain

### DIFF
--- a/app/repository/event.py
+++ b/app/repository/event.py
@@ -1,6 +1,5 @@
 """Operations on the repository for the Family entity."""
 
-import copy
 import logging
 import os
 from datetime import datetime
@@ -230,10 +229,6 @@ def update(
         _LOGGER.error(f"Unable to find event for update {import_id}")
         return False
 
-    metadata = original_fe.valid_metadata
-    if isinstance(metadata, dict):
-        metadata = copy.deepcopy(metadata)
-
     result = db.execute(
         db_update(FamilyEvent)
         .where(FamilyEvent.import_id == original_fe.import_id)
@@ -241,7 +236,7 @@ def update(
             title=new_values["event_title"],
             event_type_name=new_values["event_type_value"],
             date=new_values["date"],
-            valid_metadata=event.metadata or metadata,
+            valid_metadata=event.metadata,
         )
     )
 


app/repository/event.py
```diff
@@ -85,18 +84,17 @@ def _dto_to_event_dict(dto: EventCreateDTO, event_metadata) -> dict:
         "title": dto.event_title,
         "event_type_name": dto.event_type_value,
         "status": EventStatus.OK,
-        "valid_metadata": event_metadata,  # TODO: Fix as part of PDCT-1622
+        "valid_metadata": dto.metadata.model_dump(exclude_none=True),
```
annaCPR, line 87: Looking for comments/suggestions on this. Because the new litigation event taxonomy has a "description" field but others don't I'm having to exclude any fields that have None values because they fail validation against their taxonomy and we don't want to be saving events where the metadata has a key of 'description' with the value of None. This feels a bit brittle as we have to remember to do this in the right places but I couldn't think of a better way.

app/repository/event.py
```diff
@@ -250,7 +241,7 @@ def update(
             title=new_values["event_title"],
             event_type_name=new_values["event_type_value"],
             date=new_values["date"],
-            valid_metadata=event_metadata or metadata,
+            valid_metadata=event.metadata.model_dump() or metadata,
```
annaCPR, line 244: See comment above

app/service/event.py
```diff
@@ -119,18 +115,15 @@ def create(
...
     )
 
-    event_metadata = create_event_metadata_object(
-        db, family.corpus_import_id, event.event_type_value
-    )
     metadata_service.validate_metadata(
         db,
         family.corpus_import_id,
-        event_metadata,
+        event.metadata.model_dump(exclude_none=True),
```
annaCPR, line 121: See comment above

app/service/event.py
```diff
@@ -177,13 +170,10 @@ def update(
...
     app_user.raise_if_unauthorised_to_make_changes(user, entity_org_id, import_id)
 
-    event_metadata = create_event_metadata_object(
-        db, family.corpus_import_id, event.event_type_value
-    )
     metadata_service.validate_metadata(
         db,
         family.corpus_import_id,
-        event_metadata,
+        event.metadata.model_dump(exclude_none=True),
```
annaCPR, line 176: See comment above

app/repository/event.py
katybaulch: I wonder whether we've overcomplicated things by making `metadata` a pydantic typed model rather than just using 'Json' type like we do for documents? I think using 'Json' type means we don't need to do this, and the validation can be run in the [db_client](https://github.com/climatepolicyradar/navigator-db-client/blob/fac0b04e70ecf1845fc729cabfaec6954319aeea/db_client/functions/metadata.py#L93) against the schema as we'd expect. What are your thoughts?

annaCPR: Yeah that's a good point. I had good intentions 😆 Let me go back to the simplified way.

app/repository/event.py
```diff
@@ -240,8 +233,6 @@ def update(
     metadata = original_fe.valid_metadata
     if isinstance(metadata, dict):
```
katybaulch, line 234: question: do we still need this?

annaCPR: We probably don't actually now that you mention it. Will remove

PR #362, opened 2025-07-09
feat: add latest to ECR push

jamesgorrie:
# Description

- adds `latest` to the ECR release

Any tidy up can happen when we have this process running e2e.


katybaulch (approved):
Lgtm! Do we still need the deploy to staging job if we move that into the merge to main?? 🤔

PR #375, opened 2025-08-20
fix: prod => production

jamesgorrie:
# Description

[As per the RFC](https://github.com/climatepolicyradar/architecture-hub/blob/main/adrs/2025-04-28-ADR-006-environment-naming-conventions.md) - we should be using `production` over `prod`.


katybaulch (approved):
lgtm

